PR #9, opened 2020-02-11
Fix in production build script

framefactory:
- Fixes a bug in the Webpack/Typescript build pipeline which causes the production build to fail, switches from "awesome-typescript-loader" to "ts-loader" for Typescript loading/compilation.
- Minor fix with respect to properly detect the global Google Analytics function.


ghalusa:
Thanks @framefactory! We ran tests, and they're good to go.

PR #179, opened 2022-12-16
Observe changes to document attribute

### LINKED ISSUE #177: Switching models in explorer
I'm trying to write some custom navigation around Voyager Explorer that allows me to swap out the document being viewed. Previously, my buttons all had `href="?document=foo.json"` attributes, effectively reloading the whole viewer every time. I want to make this a bit more smooth, though, and reuse the page that I already have open.

The way that seems the most obvious is to change the document property on the application:

```javascript
explorer.application.props.document = "foo.json";
explorer.application.evaluateProps();
```

This successfully changes the document, but notably the title does not update with the contents of the new document. Looking through the code for `evaluateProps()`, it occurred to me that I could do it manually but disable document merging:

```javascript
explorer.application.loadDocument("foo.json", false);
explorer.application.postLoadHandler(explorer.props);
```

This causes the title to update and in the dev tools I can see that the correct model path is accessed, but the scene goes black. The tools and overlay buttons all function properly. Clearly there's something about the viewer state here that I don't understand.

I know this isn't part of the official API, so I understand if what I'm trying isn't possible. I can hackily get around this by creating a new `voyager-explorer` element from scratch and replacing the existing one in the DOM, but that seems inelegant and it would be nice to avoid it.

csparker247:
Implement changing the document by setting the attribute.

Closes #177

### DIFF
--- a/source/client/applications/ExplorerApplication.ts
+++ b/source/client/applications/ExplorerApplication.ts
@@ -247,6 +247,13 @@ Version: ${ENV_VERSION}
             });
     }
 
+    reloadDocument()
+    {
+        this.documentProvider.removeActiveDocument();
+        this.documentProvider.createDocument(documentTemplate as any);
+        this.evaluateProps();
+    }
+
     loadModel(modelPath: string, quality: string)
     {
         return this.documentProvider.appendModel(modelPath, quality);
--- a/source/client/ui/explorer/MainView.ts
+++ b/source/client/ui/explorer/MainView.ts
@@ -74,7 +74,7 @@ export default class MainView extends CustomElement
 {
     application: ExplorerApplication = null;
 
-    static get observedAttributes() { return ['root']; }
+    static get observedAttributes() { return ['root', 'document']; }
 
     constructor(application?: ExplorerApplication)
     {
@@ -169,6 +169,10 @@ export default class MainView extends CustomElement
 
         if(this.application && name === "root") {
             this.application.props.root = this.getAttribute("root");
+            this.application.evaluateProps();
+        }
+        else if(this.application && name === "document") {
+            this.application.props.document = this.getAttribute("document");
             this.application.evaluateProps();
         }
         else if(this.application && name === "controls") {
--- a/source/client/ui/explorer/MainView.ts
+++ b/source/client/ui/explorer/MainView.ts
@@ -125,8 +125,6 @@ export default class MainView extends CustomElement
         this.attachShadow({mode: 'open'});
         const shadowRoot = this.shadowRoot;
 
-        this.arManager.shadowRoot = shadowRoot;
-
         // add style
         var styleElement = document.createElement("style");
         styleElement.innerText = styles;
@@ -153,6 +151,7 @@ export default class MainView extends CustomElement
     {
         this.fullscreen.fullscreenElement = this;
         this.viewer.rootElement = this;
+        this.arManager.shadowRoot = this.shadowRoot;
     }
 
     protected disconnected()
@@ -165,18 +164,21 @@ export default class MainView extends CustomElement
 
     attributeChangedCallback(name: string, old: string | null, value: string | null)
     {
+        const app = this.application;
         super.attributeChangedCallback(name, old, value);
 
-        if(this.application && name === "root") {
-            this.application.props.root = this.getAttribute("root");
-            this.application.evaluateProps();
+        if(app && name === "root") {
+            app.props.root = this.getAttribute("root");
+            app.reloadDocument();
+            this.connected();
         }
-        else if(this.application && name === "document") {
-            this.application.props.document = this.getAttribute("document");
-            this.application.evaluateProps();
+        else if(app && name === "document") {
+            app.props.document = this.getAttribute("document");
+            app.reloadDocument();
+            this.connected();
         }
-        else if(this.application && name === "controls") {
-            this.application.enableNavigation(value);
+        else if(app && name === "controls") {
+            app.enableNavigation(value);
         }
     }
 


source/client/components/CVDocument.ts
```diff
@@ -183,8 +183,8 @@ export default class CVDocument extends CRenderGraph
 
         if (!mergeParent) {
             this.clearNodeTree();
-            this.ins.title.setValue(null);
         }
+        this.ins.title.setValue(null);
```
csparker247, line 187: This works for updating the title but certainly seems like it could have unintended side effects. Maybe I should check if the new document has a title before clearing this value?

csparker247: Of course, one alternative might be to leave this as it originally was and modify the call from `evaluateProps` to not merge the parent. The issue I have there is that the entire viewer disappears, presumably because of `clearNodeTree()` here.

csparker247: With a little more time, I've tracked this down to where the update should actually happen in `CVDocument`. This alone is also enough to force the title to update, but is likewise prone to unintended side effects:
```diff
# CVDocument.ts, line 293
-        if (event.add && !propTitle.value) {
+        if (event.add) {
             meta.once("load", () => {
                 this.titles = meta.collection.get("titles") || {};
```

I can see a number of ways around this at this point, but I'm not sure what the safest solution is. I'm inclined to unset `title` in `ExplorerApplication.evaluateProps()`. This would seem to localize the effects to only Explorer where changing the document property is (perhaps) expected to change the scene:
```diff
# ExplorerApplication.ts, line 325
         if (props.document) {
             // first loading priority: document
             props.document = props.root ? props.document : manager.getAssetName(props.document);
+            this.documentProvider.activeComponent.ins.title.setValue(null);
             this.loadDocument(props.document, undefined, props.quality)
```
I'm going to push this approach for now. If there's a better way, let me know what you prefer.

gjcope:
Are you able to share an example showing the title issue? With your original commit observing the document attribute I'm able to swap scenes and I'm not seeing a title issue.

csparker247:
Sure! 

[PHerc0118Cr01.json.txt](https://github.com/Smithsonian/dpo-voyager/files/10261478/PHerc0118Cr01.json.txt)
[PHerc0118Cr01_IR.json.txt](https://github.com/Smithsonian/dpo-voyager/files/10261477/PHerc0118Cr01_IR.json.txt)

gjcope:
If I'm understanding it correctly, it looks like you aren't successfully loading the new document due to errors with asset paths. If this is the case, it may be a need for more graceful error state handling. Do you have an example where the scenes successfully load but the title doesn't change?

csparker247:
That loading error is it failing to load the glb from the document because it's being hosted on a webserver and I'm running the app locally (cross origin request). That's expected. The documents themselves are local and are getting switched correctly, as evidenced by the path to the glb changing between doc-a and doc-b. 

Here's a video when I run the same thing on the most recent commit of my branch:
https://user-images.githubusercontent.com/1434526/208507378-8757e177-b3f6-4bed-b5a2-6a798bf28255.mp4

In both videos, notice the value of the title element. In this video, I start with `document` unset, so it displays `Missing Title`. I then set the attribute to `doc-a.json`, which changes the title to `P.Herc. 89, Cr. 01`. I then set the attribute to `doc-b.json` which changes the title to `P.Herc. 89, Cr. 01 (IR)`. In the first commit, the title does not update when I do this last step.

I'll spin up a dev site for each of these commits so you can see it with the actual data.

csparker247:
Here you go:
- Using the most recent commit: http://infoforest.cs.uky.edu/voyager/change-model/
- Use the first commit: http://infoforest.cs.uky.edu/voyager/change-model-first-commit/

The only code being used to apply the change is:
```javascript
    // Load the doc in explorer
    let explorer = document.getElementById("explorer");
    explorer.setAttribute("document", doc);
```

gjcope:
Thanks! I'd like to dive into this a bit more. I'm not able to recreate with my own documents, so I will give it a go with yours.

csparker247:
Yeah, there's definitely something weird going on. I added the chair demo asset to the two links above. When the chair didn't have a meta assigned to the scene, the switching worked as expected, but would display Missing Title for the chair. I've now updated the chair document with a title, and switching to the chair works but switching back to my other models doesn't...

This PR branch should be setup so that you can push to it. I'm not going to commit anymore until I hear from you, so feel free to use it if you want.

gjcope:
There's something more fundamental at issue so I'm going to address that and then merge your PR. I worked backwards from my scene file (not showing the issue) to your bare-bones one and it looks like the issue is there not being a camera defined in the scene. If I add one to your scene files it works as expected. But this shouldn't be a requirement, so I'll continue investigating.

gjcope:
Apologies for the delay @csparker247 , finally was able to get back to this. I think the root of the problem is here:
https://github.com/Smithsonian/dpo-voyager/blob/833cc1e35999471d24c5fc223adbfcc8c57daaa9/source/client/applications/ExplorerApplication.ts#L238
If merge is not defined and there are no lights or no cameras (as in your examples) it assumes a merge. I think the fix is to explicitly set merge to false in evaluateProps:
https://github.com/Smithsonian/dpo-voyager/blob/833cc1e35999471d24c5fc223adbfcc8c57daaa9/source/client/applications/ExplorerApplication.ts#L328
If evaluateProps is being executed, we are either loading or reloading - I can't think of a scenario where merging would be warranted here. Any thoughts? If it seems appropriate, can you try that and confirm it works for you? Thanks!

csparker247:
No worries!

I'll give it a try, but I think I tried something like this back in the original issue. It loaded the doc but the scene went black and didn't show anything but the new title. I'll let you know if that's what happens here.

gjcope:
Good point, sorry I missed that. I think there are two things happening. The title not changing was due to the merging. Without the merging, it is swapping scenes correctly but not rendering correctly due to the lack of camera in the scene. I will take another look.

csparker247:
Sounds good. I couldn't quite tell what was getting deleted before. My thought had been the whole rendering context, but camera makes more sense given that I'm relying on the default camera. 

 I'm AFK until Tuesday, but I'll roll back to the first commit then since it looks like my second change won't be necessary if we explicitly disable merging.

gjcope:
Took me minute to dig through the loading code but it looks like there is a bit more to it.

At app load, a default document is created with the template. 

When a scene document is loaded, if it has lights/cameras it replaces the default doc. If not, it's assumed that you want to merge with the default doc to get the default lights and camera. 

This works great with a single scene, but once you attempt to load another scene without lights/cameras at runtime you are now merging with the prior document rather than the default template.

Disabling the merge fixes this, but then we are also getting no lights/camera from the default template which results in the black screen.

sdumetz:
I'm working on scenes reload too and there is another problem : Even using scenes with lights and camera, forcing `merge = false`, there is some memory leak. The document's data seems to be kept in memory. Even if I switch back and forth on the same few documents, it uses more memory on every reload.

I've traced it down to the `CVDocument.clearNodeTree()` (apparently) not releasing all resources properly but maybe you have some insights on where and why the leak happens?

I've tried to recreate the document from scratch using `this.documentProvider.createDocument(documentTemplate as any);` but it fails too because some nodes are not recreated properly.

gjcope:
> I'm working on scenes reload too and there is another problem : Even using scenes with lights and camera, forcing `merge = false`, there is some memory leak. The document's data seems to be kept in memory. Even if I switch back and forth on the same few documents, it uses more memory on every reload.
> 
> I've traced it down to the `CVDocument.clearNodeTree()` (apparently) not releasing all resources properly but maybe you have some insights on where and why the leak happens?
> 
> I've tried to recreate the document from scratch using `this.documentProvider.createDocument(documentTemplate as any);` but it fails too because some nodes are not recreated properly.

@sdumetz Do you mind opening a separate issue for this? Thanks!

csparker247:
Alright, I've reset back to the first commit and pushed a new commit with merging disabled in `evaluateProps()`.

> When a scene document is loaded, if it has lights/cameras it replaces the default doc. If not, it's assumed that you want to merge with the default doc to get the default lights and camera.

Woof, this is a bit of a mess to sort out. I think we could hack around this, but I'm having trouble parsing what the appropriate placement and conditional for this might be. This seems to be the approach that most closely resembles the initial conditions of the viewer, right?
```diff
         return this.assetReader.getJSON(documentPath)
             .then(data => {
+                if(merge === false) {
+                    this.documentProvider.createDocument(documentTemplate as any)
+                }
                 merge = merge === undefined ? !data.lights && !data.cameras : merge;
                 return this.documentProvider.amendDocument(data, documentPath, merge);
             })
```

That said, it would be nice for my use case to actually keep the current camera and lighting when the model is swapped. But needs must.

gjcope:
Was thinking along the same lines, but possibly moving the document reset. Maybe clearing out the current scene when handling the attribute change could work since this seems like a pretty specific scenario. If I've changed the document, I'm looking to replace the scene rather than merge. I'll be able to test something out later this week. Thanks for the help!

csparker247:
Awesome! That sounds like a solution with fewer unintended consequences. Let me know if you need me to do anything.

gjcope:
I think we have a candidate! If you have a chance, take a look at this branch: https://github.com/Smithsonian/dpo-voyager/compare/master...dev-reload

The most recent commit adds a document reload function to the app that clears the old document and loads in the template. This should accommodate files with or without lights/cameras the same way they are on initial load.

I also checked memory usage and it looks pretty good to me. There are a few bits that should be tracked down and released but I'm not seeing anything significant. Please let me know if you experience otherwise.

If it looks good I would love if you updated here so I can merge and you get the credit for all the investigative heavy lifting! ;)

csparker247:
Excellent! It all seems to work in my test site. The updated code has been pushed. Thanks for the help (and the credit 😄).

gjcope (approved):

gjcope:
Done! Thanks for your help @csparker247 !

PR #180, opened 2023-01-26
update articles and annotations languages when UI lang is not the default

sdumetz:
when UI locale is not the default value, all articles titles are undefined.

The source is that `CVReader.onLanguageComponent()` was called once, before the articles are loaded (maybe it's a race condition and doesn't happen every time, didn't investigate).

My change does two things : 
- set `CVReader.language` to not throw an error if `CVLanguageManager` doesn't exist yet. I belive it is the intended behavior as it's generally discouraged to have properties getters throw errors and I couldn't find anywhere it was supposed to.
- call `CVReader.updateArticles()` if the language manager is ever loaded before the meta component. Prevents an hypothetical race condition
- if articles are updated and the language is loaded, set the language using `CVReader.updateLanguage()`

I made the minimal viable change but there is more to do on this subject. I'll make more pull requests when I can.

### DIFF
--- a/source/client/components/CVAnnotationView.ts
+++ b/source/client/components/CVAnnotationView.ts
@@ -388,7 +388,11 @@ export default class CVAnnotationView extends CObject3D
 
     fromData(data: IAnnotation[])
     {
-        data.forEach(annotationJson => this.addAnnotation(new Annotation(annotationJson)));
+        data.forEach(annotationJson => {
+            let annotation = new Annotation(annotationJson);
+            annotation.language = this.language.outs.language.value;
+            this.addAnnotation(annotation);
+        });
         this.emit<ITagUpdateEvent>({ type: "tag-update" });
     }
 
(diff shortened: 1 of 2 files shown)


sdumetz:
Made very similar changes for annotations, I thought they should be together.

For annotations, `this.language` seems always defined. It made me question my assumptions, maybe there is some deeper bug here? Let me know if you think it deserve more investigations.

gjcope:
Hi @sdumetz , thanks for highlighting the inconsistent behavior. I'd like to investigate a little further to make sure we are consistent on all fronts.

gjcope:
Hi @sdumetz finally had a chance to look at this. Noticed there was also an issue with UI tooltips in this case and found that the solution for all of the issues is even simpler. There's a bug in the language manager.

Remove this line: https://github.com/Smithsonian/dpo-voyager/blob/1a1e7185e3a577be42d903fbe7d8bdcfd7fbff82/source/client/components/CVLanguageManager.ts#L113
and everything should work as expected. Please let me know if it works for you as well.

sdumetz:
Works like a charm, thanks!

gjcope:
Great, thanks for highlighting the issue. Fix will be included in the next release.

Issue #41, opened 2021-01-22
Multi-language support for Story UI


gjcope:
Complete language support by adding a language manager for the story UI. Point all language-affected tasks to this manager.

ndejardin425:
Just hopping on the train, but from what I've understand in the langagueManager, there is 

- activeLanguage -> language for the content (articles, annotations, ...) -> can set using lang parameter 
- uiLanguage -> language for the interface -> can be set using uiLang parameter (but is itdocumented in the API Overview ?)
- primarySceneLanguage -> ???

If I want everything to be in french in Story mode ? do I have to just do                     
`<voyager-story lang="fr" uiLang="fr` ?

and for exemple, in Property (like in my new PropertyTags) should we use language.getLocalizedString or language.getUILocalizedString ?

Inao0:
Hello,

_primarySceneLanguage_ is the default language of the scene. If you want more context, you could have a look at the discussions in #359.

If you use  `<voyager-story lang="fr" uiLang="fr ?` you will see the scene in French and story ui in French, but the scene primary scene might not be set properly. On using the reset button, the scene will go back to its primary scene language which is set according to `DEFAULT_LANGUAGE` (which is English by default) unless overriden by setting another primary scene language using the capture task.

_getLocalizedString_ gets the string in the active language (the language the scene is currently displayed in). It is mostly used in Explorer UI. 
_getUILocalizedString_  gets the string in the uiLanguage. This is probably the one you want for an element of Story ui.

PR #184, opened 2023-02-08
Story UI localization + French translations

sdumetz:
So I've combined some related changes : 
- added `CVLanguageManager` to `TaskBar` the same way it is currently added in `ToolBar`.
- added a conditional log in CVLanguageManager to find strings that are not translated when not in English and not in development mode.
- all strings that were in `string.resources.en.json` were translated in a new `string.resources.fr.json` file. I also added missing strings that appeared from 956926eaa5e56532930a9cc815a06ab95b65ecbc

It's of course possible to split any of the above in separate requests if needed.


gjcope (approved):
This looks great. The LanguageManager logging is a nice idea. TaskBar localization is one step  towards #41 but appreciated! Thanks.

PR #186, opened 2023-02-13
Add a dialog to define new model derivatives

sdumetz:
So I prototyped something that you might want to have:

I wanted to manage my 3D files in the way of **Standalone** mode but using the normal backend.

As a first step, I retrofitted the `ImportMenu.show()` to be called when the user double-clicks on a model in the `AssetTree`,  like what happens with the `ArticleEditor`.

It allows users to define new derivatives for a model or create a new model in the scene.

In my mind, the finished feature would contain the following : 

- Upon uploading a new model (drag & drop into asset-tree), `ImportMenu.show()` is shown to assign it to the scene
- An UI element should be created to un-assign derivatives to a model, to provide a way to correct mistakes Probably something like a double-click on the model in the **Navigator** tab?

Ideally I'd also like to make the `CVStandaloneFileManager` general-purpose to allow users to drag-drop into the main frame. The methods designed to get/set files could be isolated into another class, possibly simplifying the conditionals in `CVMediaManager`, `CVAssetWriter`, and others.

As I understand it those features are not required to your workflow but would you be open to merge them if done properly? Or do you think it would break how you expect the application to behave in a way I didn't understand?

### DIFF
--- a/source/client/applications/StoryApplication.ts
+++ b/source/client/applications/StoryApplication.ts
@@ -42,6 +42,7 @@ import NVoyagerStory from "../nodes/NVoyagerStory";
 import MainView from "../ui/story/MainView";
 import CVTaskProvider, { ETaskMode } from "../components/CVTaskProvider";
 import CVStandaloneFileManager from "client/components/CVStandaloneFileManager";
+import CVModelEditor from "client/components/CVModelEditor";
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -189,6 +190,8 @@ export default class StoryApplication
             props.dragdrop = true;
             app.createComponent(CVStandaloneFileManager);
         }
+
+        app.createComponent(CVModelEditor);
 
         this.mediaManager.rootUrl = this.assetManager.baseUrl;
 
--- a/source/client/applications/StoryApplication.ts
+++ b/source/client/applications/StoryApplication.ts
@@ -56,6 +56,8 @@ export interface IStoryApplicationProps extends IExplorerApplicationProps
     mode?: string;
     /** When set to true, application displays additional expert level tools. */
     expert?: boolean;
+    /** When set to true, application supports dragging and dropping files. */
+    dragdrop?: boolean;
 }
 
 /**
@@ -153,6 +155,7 @@ export default class StoryApplication
         props.referrer = props.referrer || parseUrlParameter("referrer");
         props.mode = props.mode || parseUrlParameter("mode") || "prep";
         props.expert = props.expert !== undefined ? props.expert : parseUrlParameter("expert") !== "false";
+        props.dragdrop = props.dragdrop || false; 
 
         // If in standalone mode, remove root and document params that may be present
         const modeText = props.mode.toLowerCase();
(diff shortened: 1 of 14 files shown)


gjcope:
Hi @sdumetz - This is specifically outside of our workflow since our scenes are produced directly from https://github.com/Smithsonian/dpo-cook with all derivatives. That's also one of the reasons we created the standalone version for those not using Cook (and also to remove the WebDAV requirement). But we have heard from a couple of collaborators that this would be useful so we are definitely open to considering it, we just have to think about how it fits in our regular workflow.

Also tagging @kgschoueri here as her group is interested in something similar.

gjcope:
@sdumetz I had some time to work on this. Please see the dev-dragdrop branch. It's still a WIP and I would appreciate any help/input. I've moved the actual drag/drop out of standalone and added an attribute to enable it. Also stubbed out a place to consolidate initial file handling independent of mode (https://github.com/Smithsonian/dpo-voyager/blob/3c73694a771fa84faa3c6bc4120630c29e4e94a5/source/client/components/CVMediaManager.ts#L96) I think once generalized drag/drop is in place it makes sense to improve the UX with workflows like what you propose here.

sdumetz:
I like what you did there.

I have ideas for the UX, I'll try to make a more advanced prototype based on your work in a few days.

The one thing I don't know how to change is the `<sv-asset-panel>`. Currently it filters which files to show depending on "mode". However we can't expect a user who just dropped the wrong file (say, a model) to switch to Expert Mode to be able to delete it.

I see two easy workarounds that would work well for me :
 - unfiltered asset-tree everywhere
 - unfiltered asset-tree in Edit mode and restrict drag&drop to this mode

However I'm not sure if it's OK for you? You might want to keep most files hidden outside of expert mode? Maybe tie the filter to the `dragdrop` flag you just added?

sdumetz:
So I got something. It's not as straightforward as I'd like it to be but it's a start.

you can check it out here : https://github.com/Holusion/e-thesaurus/tree/model_edit

It would require small changes to `ff-ui` and `ff-scene` to work (PRs on the way), then I can backport it here.

It implements the following features : 

- drag & drop from `ff-asset-tree` to the viewport opens the `ImportMenu` Popup to add a model/derivative to the scene
- The derivatives panel allows suppression of derivatives and adding new derivatives

The drag & drop works great but is not very discoverable and as such, not a great feature. But it's a really small change. 

I'm not really pleased with the way I got the derivatives Panel to work, but it's the simplest way to do derivatives management that I could think of.

gjcope:
Great, thanks. Stuck on another project this week but will take a look asap. There is also still work to be done to filter uploads and manage derivatives in non-standalone mode to replicate the existing standalone functionality. I think this is important for consistency and I also like having the viewport as a big target for drag/drop.

gjcope:
drag/drop branch is updated with full function equal to standalone mode. Code has been consolidated and cleaned up a bit, though it could probably be separated out from MediaManager.

I could use some extra eyes/testing for your use cases. I'm sure there are some issues with clearing things out when drag/dropping a full scene package over an existing one, but other than that haven't noticed anything significant.

kgschoueri:
this sounds great also for our needs! Will the new functionality be available in the next release? When are you planning for that to happen?

sdumetz:
@gjcope I've rebased on the new commits.  I didn't see any issues, but can't say I tested it thoroughly as I don't use full scenes upload (`.svx.json` files), only derivatives of existing objects or new objects to add to an existing scene.

From my point of view, this PR _could_ be deployed once https://github.com/Smithsonian/ff-scene/pull/4 is merged and updated in voyager (ideally, https://github.com/Smithsonian/ff-ui/pull/2 too, but not required).

Let me know if changes are needed.

gjcope:
> this sounds great also for our needs! Will the new functionality be available in the next release? When are you planning for that to happen?

@kgschoueri Yes, we would love for you to test prior to release as it isn't part of our internal workflow. The goal is to include it in the release at the end of the month, but you can test it immediately with the dev-dragdrop branch.

PR #187, opened 2023-02-15
responsive header in story mode

pakrettt:
The header in story mode overflow in small devices.

So I folded it by adding a scrollbar and displayed only the icons.

### DIFF
--- a/source/client/ui/story/styles.scss
+++ b/source/client/ui/story/styles.scss
@@ -88,6 +88,8 @@ ff-tab-header, ff-dock-panel-header {
   background-color: $color-background-dark;
   border-bottom: 3px solid $color-background-darker;
 
+  overflow: auto;
+
   .ff-group {
     align-items: stretch;
   }
@@ -96,6 +98,10 @@ ff-tab-header, ff-dock-panel-header {
     height: 28px !important; 
     margin: 8px;
     align-self: center;
+
+    @media (max-width: 1080px){
+      display: none;
+    }
   }
 
   .sv-mode {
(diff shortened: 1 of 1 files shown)


gjcope:
We have always considered Story to be pretty unusable on small display devices due to the complex UI and required space to work with all the features. A responsive header sounds like a step in that direction, but even with that change have you found it to be usable on small devices? How small?

sdumetz:
Without this change I have the header causing an ugly overflow as soon as I go below ~950px width.

The rest of the interface is clearly not good for mobile/small screen editing (nor does it need to be) but it remains useable for small changes even with something as small as an iPhone, which we like to be able to do when a scene needs to be fixed ASAP and we're not in the office.

There might be some more low hanging fruits there that we will contribute when we can.

gjcope:
That's fair and makes sense. With that in mind, I really like the the overflow update and the scrolling. 

I think the idea of hiding things like the logo and mode icons once the menu overflows is good but should not be based on absolute window size. A feature we are still working on is having task modes that are specific to certain user types, so they only see the features that are relevant. For instance, add &mode=qc to the end of your url and you will get a menu with less options which then overflows at a much smaller width.

source/client/ui/story/styles.scss
```diff
@@ -96,6 +98,10 @@ ff-tab-header, ff-dock-panel-header {
     height: 28px !important; 
     margin: 8px;
     align-self: center;
+
```
gjcope, line 101: Please remove - per discussion comment, this makes functional sense but should not be tied to  a specific width.

gjcope (changes requested):
Please remove attributes tied to 'max-width'. Changing the display of these pieces makes sense, but should not be tied to width since the taskbar is variable depending on mode. Maybe an IntersectionObserver?

gjcope:
I would love to include this style change in the next release if the change request can be made so you can get credit.

gjcope:
@pakrettt @sdumetz Please update or I will close and include style change with a mention as credit.

gjcope:
No response so closing and added overflow update with tag @pakrettt

PR #188, opened 2023-02-21
use native Material.copy method

sdumetz:
So I noticed the `UberPBRMaterial` was missing the emissive map from one of my models.

I checked why and found `UberPBRMaterial.copyStandardMaterial()`was simply not copying it.

Instead of just adding it I wondered if other properties were missed in the copy and it turns out a whole bunch of them were.

The native `Material.copy()`method from THREE (code [here](https://github.com/mrdoob/three.js/blob/dev/src/materials/Material.js#L404)) seems to do the job, with the added benefit of being maintained upstream.

I tested it on my models and it looked right, fixing the missing emissive and not breaking anything. i'd like to know if it breaks anything for you, or if  you would rather only include the required properties in `copyStandardMaterial`.

### DIFF
--- a/source/client/io/ModelReader.ts
+++ b/source/client/io/ModelReader.ts
@@ -119,7 +119,7 @@ export default class ModelReader
 
                 // copy properties from previous material
                 if (material.type === "MeshStandardMaterial") {
-                    uberMat.copyStandardMaterial(material);
+                    uberMat.copy(material);
                 }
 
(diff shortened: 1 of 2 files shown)


gjcope:
It seems to work fine for me. We definitely aren't supporting all the map types in various other ways, but I don't see a problem with simplifying in this way and dealing with other map types as a business case for them arises. This shouldn't add any overhead.

gjcope (approved):

PR #198, opened 2023-03-23
add button in tinyMCE to edit raw html code

sdumetz:
I have some user that want to get more creative with their article's layout.

The [Code Plugin](https://www.tiny.cloud/docs/tinymce/6/code/) shipped with tinyMCe seems OK enough for them.

I've tested some basic things like two-columns layout, image styling. It's obviously a "use at your own risk" feature, but any required security sanitization should be done on the server anyways.

### DIFF
--- a/source/client/ui/story/ArticleEditor.ts
+++ b/source/client/ui/story/ArticleEditor.ts
@@ -34,6 +34,7 @@ import 'tinymce/plugins/link';
 import 'tinymce/plugins/lists';
 import 'tinymce/plugins/image';
 import 'tinymce/plugins/media';
+import 'tinymce/plugins/code';
 
 /* Import content css */
 import contentUiCss from '!!raw-loader!./editor_css/content.ui.min.css';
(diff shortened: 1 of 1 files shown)


kgschoueri:
I am also interested in some added capability for the html article, specifically a way to add hspace and a caption under the image without having to hack the html file outside of the Voyager Story interface.

kgschoueri:
@gjcope that is great, thank you! I will keep these in mind but for me this was a major thing that will be great to have directly in the UI

gjcope:
> @gjcope that is great, thank you! I will keep these in mind but for me this was a major thing that will be great to have directly in the UI

Editing the html directly is important for you to have in the UI?

sdumetz:
@gjcope we don’t normally expose expert mode so merging it only for this mode will still require a patch on our side unless we make some UI to "quick-switch" (thinking about it, will let you know).

To provide more feedback, we have been using it for a few weeks and it’s quite limited : the code is parsed back into tinyMCE **before** save, filtering out some properties. So columns are OK but some things like responsive video embeds are not possible RN for some reason.

so I’ll probably look for a better solution in a few weeks/months.

In the meantime, there is no need to merge it if no one is using it. It’s a small isolated patch so we don’t really care if it’s upstreamed or not : there is no real risk it would make future contributions more difficult.

gjcope:
@sdumetz Got it, thanks for the context. I think I'll close it for now, but happy to consider a little more restrictive solution down the road. Our issue is that people *would* use it, and do weird things. ;)

PR #199, opened 2023-03-23
allow upload of medias in tinymce

sdumetz:
The image widget has an "upload" panel but that's not available in the "medias" widget.

If `file_picker_callback` is defined, an icon appears next to each media inputs to allow uploading something from the client's computer.

Works fine for me though I couldn't find a good way to signal the in-progress upload to the user in this context. I used `Notification.show` but it's far from ideal.


gjcope:
We currently can't support this feature on our back end, so opening it up to our users would cause problems. I've opened an issue (#209 ) to track this, so we can revisit your solution when it will work for our users. As always, thanks for the PR!

PR #204, opened 2023-04-12
accept PLY and OBJ files in ingestFiles

sdumetz:
OBJ and PLY files were rejected from injestFiles, though theoretically supported.

As they are transmitted as Blob URL to `GeometryReader` in standalone mode, it required to pass the original extension as an additional optional parameter to work, with some other minor changes.

I now realize I rewrote most of `CVMediaManager.ingestFiles()` for no good reason. I find it more readable this way but  I can revert those changes if you prefer.

### DIFF
--- a/source/client/components/CVAssetReader.ts
+++ b/source/client/components/CVAssetReader.ts
@@ -107,7 +107,7 @@ export default class CVAssetReader extends Component
     async getGeometry(assetPath: string): Promise<THREE.BufferGeometry>
     {
         const url = this.assetManager.getAssetUrl(assetPath);
-        return this.geometryLoader.get(url);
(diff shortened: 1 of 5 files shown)


sdumetz:
I agree with you that there is no rush in merging this until someone intends to publish serious work that requires it: It's more or less worthless for now because it won't map textures properly.

However I will be working on point clouds in the near future (funding is not yet secured for this feature, but should be soon). It works as a prototype with very little changes and will require first-class PLY support (feature branch for reference : https://github.com/Holusion/eCorpus/pull/30 ).

I'll submit a PR when my work on points cloud is done. We could discuss this further then.

gjcope:
That's great news and definitely a reason to promote PLY. Looking forward to seeing it.

PR #211, opened 2023-05-04
fix a race condition in CVLanguageManager

sdumetz:
- load the UI in french (using a query string or custom-element attribute)
- load a scene set-up to be in english (`setups[0].language.language = "EN"`in the scene file)

The language widget shows the scene as **english**, but it's (sometimes) displayed in french.

It is the result of a race condition: The french dictionary is requested before the document, then if it loads after the setup has been parsed, it changes back the output language.

This commit fixes it.


gjcope (approved):
I couldn't reproduce the issue, but that could just be a difference in load times. The fix makes sense. Good catch!

gjcope:
Thanks for the PR!

PR #215, opened 2023-05-22
support light creation

sdumetz:
This PR touches a lot of things :
- add a button to add/remove nodes in node tree
- add a popup menu to create new nodes (only lights supported, but I intend to reuse it with https://github.com/Smithsonian/dpo-voyager/pull/186 ).
- add support for every light source supported in THREE.
- use azimuth/elevation to configure `CVDirectionalLight` nodes

Everything should be backward compatible with existing scenes.

I've put the whole work here, but it can obviously be broken down if some parts are problematic. In particular, support for `RectAreaLight` is currently suboptimal (missing helper, positionning is bad).

Additionally, it uses my fork of [ff-scene](https://github.com/Smithsonian/ff-scene) to provide the new light types. 
 https://github.com/Smithsonian/ff-scene/pull/5 should be merged first.

In my dev deployment, I replaced the default 4 points directional with a `HemisphereLight` and one `DirectionalLight` (see [https://github.com/Holusion/eCorpus/commit/6f3b32afb005e4406fe3378b08735d094ca50770) and I think it looks better for most scenes, especially with Tone Mapping enabled. In any case, the additional configurability is really nice.

Let me know if you want to restrict some options to expert mode, or any other change that may be necessary.

### DIFF
--- a/source/client/applications/coreTypes.ts
+++ b/source/client/applications/coreTypes.ts
@@ -76,8 +76,20 @@ import NVDocuments from "../nodes/NVDocuments";
 import NVScene from "../nodes/NVScene";
 import NVNode from "../nodes/NVNode";
 import CVAudioManager from "client/components/CVAudioManager";
+import CVAmbienLight from "client/components/CVAmbientLight";
+import CVHemisphereLight from "client/components/CVHemisphereLight";
+import CVRectLight from "client/components/CVRectLight";
 
 ////////////////////////////////////////////////////////////////////////////////
+
+export const lightTypes = [
+    CVDirectionalLight,
+    CVPointLight,
+    CVSpotLight,
+    CVAmbienLight,
+    CVHemisphereLight,
+    CVRectLight,
+] as const;
 
 const types = [
     CGraph,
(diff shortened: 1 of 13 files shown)


gjcope:
Thanks for the PR @sdumetz ! It might take me a bit to get through it, but looks like a nice feature.

sdumetz:
Hi,
just a friendly reminder, if you could take time to look at this and let me know if it has a chance to get merged or if it needs to be changed.

I'd really  like to move forward with this feature.

Let me know if I can help.

gjcope:
Hi @sdumetz, thanks for the poke and apologies for the delay. I still think this is a great feature but resources are limited on our end and unfortunately lighting is a little further down the priority list. I will try to at least take a look at the UI as requested in the near future. 

One thing to take a look at is how the lighting helpers hold up with scale changes. This is something we had significant issues with and the current implement for dynamically resizing their positions is tied specifically to directional lights. I think this will probably need to be updated as well.

gjcope:
This got auto closed for some reason...

PR #217, opened 2023-06-08
fix bad value for promptVisible when readerVisible == true

sdumetz:
in scenes where readerVisible == true, `<sv-action-prompt>` doesn't get rendered and `CVOrbitNavigation.tick()` then throws an error when trying to modify the prompt's style.

I choose to always disable the prompt when the reader is visible. Another choice might be to keep it when `readerPosition !== EReaderPosition.Overlay`.

### DIFF
--- a/source/client/ui/explorer/ContentView.ts
+++ b/source/client/ui/explorer/ContentView.ts
@@ -137,7 +137,7 @@ export default class ContentView extends DocumentView
 
             if(controls && promptEnabled) {
                 const isInUse = navigation.ins.isInUse.value;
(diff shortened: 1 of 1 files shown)


gjcope (approved):
Looks good. I could see it enabled when the reader is on the right, as you mentioned, but for now I think it makes sense.

gjcope:
Right now it also disables with any UI interaction (not necessarily manipulating the model) which could be pretty easily changed by taking out isInitialLoad from that line and defaulting the style to transparent. Opinions?

PR #222, opened 2023-07-28
Mesh shadows management

sdumetz:
- allow manually forcing shadowSide to Back (otherwise its default value is tied to material.side, see [the doc](https://threejs.org/docs/#api/en/materials/Material.shadowSide)
- add toggles for `castShadow` and `receiveShadow` on models (keeping previous defaults). castShadow was always enabled (in `ModelReader`). `receiveShadow` was always disabled but it's a nice touch to have on some models.

`shadowSide` can now be toggled without having to save and reload.


gjcope:
This got auto closed for some reason...

PR #241, opened 2024-01-05
keep user-selected language when creating a new annotation

sdumetz:
When creating a new annotation, UI-locale gets silently switched back to DEFAULT_LANGUAGE, due to this line:

```
this.activeDocument.setup.language.ins.language.setValue(ELanguageType[DEFAULT_LANGUAGE]);
```
(in `CVAnnotationsTask.createAnnotation()`)

Deleting the line exposes some undefined titles values which this commit fixes. It's pretty much the same as the outdated #203, but for annotations.

Should have no impact on users working in English.

### DIFF
--- a/source/client/components/CVAnnotationsTask.ts
+++ b/source/client/components/CVAnnotationsTask.ts
@@ -36,6 +36,7 @@ import AnnotationsTaskView from "../ui/story/AnnotationsTaskView";
 import CVScene from "client/components/CVScene";
 import { ELanguageStringType, ELanguageType, DEFAULT_LANGUAGE } from "client/schema/common";
 import { getMeshTransform } from "client/utils/Helpers";
+import { IAnnotation } from "client/schema/model";
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -74,7 +75,6 @@ export default class CVAnnotationsTask extends CVTask
     constructor(node: Node, id: string)
     {
         super(node, id);
-
         const configuration = this.configuration;
         configuration.annotationsVisible = true;
         configuration.gridVisible = false;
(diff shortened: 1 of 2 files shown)


gjcope:
Hi @sdumetz thanks for the PR. This was the intended behavior, but I'm definitely open to discussion. Our thought was that annotations should always be created in the default language, with the option of adding additional language support. So if the primary application is French, I would change the default language. Is there a poor use case in this scenario? Just trying to understand the problem better.

sdumetz:
Hi @gjcope,
I understand it's logical from your perspective.
However our deployments have no concept of "default language". For example we have a service running for 2 universities in Europe where users might work primarily either in french, in Dutch  or in English. In Europe for most university, the native language AND English generally have to both be treated as "first class", with the default being chosen at the scene level rather than at the application level.

Maybe the language could be reverted to `setup.language` instead of `DEFAULT_LANGUAGE`? that would fit my use case.

My opinion is that the language switch is always a bit unsettling, even when it switches back to the "proper" (ie. scene's default) language. The change is a bit hard to notice for someone not paying close attention and I've seen a lot of users writing content in the wrong language due to this behavior.

gjcope:
Thanks for the clarification and use case. That makes sense to me, but I would like to think on it a little more and run past other collaborators frequently working with multiple languages. What I want to avoid is annotations created without the primary scene language, not necessarily primary application language as you mentioned. @kgschoueri have you seen this as an issue with your multilingual users?

sdumetz:
Closing as no longer a good solution: https://github.com/Smithsonian/dpo-voyager/pull/280 allows us to customize DEFAULT_LANGUAGE as needed.

PR #242, opened 2024-01-05
unify webpack build configs to take advantage of cache features

sdumetz:
The current build script makes webpack discover the source tree separately for each app.

Unifying the build under one config makes it ~ 3 times faster for the `all` target, while changing nothing when building individual targets.

I had to add a dependency on `copy-webpack-plugin` but that allowed me to drop `fs-extra`, removing some annoying deprecation warnings.

The package-lock.json got upgraded in the process, feel free to reset it back if you need support for older npm versions.


gjcope:
@sdumetz - Thanks! This is great but I couple of things to work out still. It looks like copy-webpack-plugin itself is not supported by older Node versions, including the one we have in the current DockerFile in the repo. That should probably be updated anyway, but I will need to figure a way to auto rebuild the image or it will break for some of our users. I'm also seeing some small differences in the build output so would like to rectify that for myself.

sdumetz:
Oh, forgot to tell you, I needed to change the application title from `Voyager-<appName> [...]` to `Voyager [...]`. because `appName` was not available in this context.

It's the only change I've seen running diff on the old vs new dist folder.

Let me know if I can help with the build process. I'm using my own Dockerfile with node_16.x to build voyager and the server-side code.

gjcope:
Merged here - https://github.com/Smithsonian/dpo-voyager/tree/dev-upgrade
It's too bad to lose the app name and the build output is a little less verbose, but the speedup is nice. The 'cleanup' flag doesn't work well for our workflow so I've commented it out. Thanks for the effort and apologies for the delay!

PR #243, opened 2024-01-09
update sass-loader: incompatible peerDep over the installed node-sass

sdumetz:
Peer dependency that caused an error under the newer npm versions:

 -`sass-loader@12.3.0` peer dependency over `node-sass@"^4.0.0 || ^5.0.0 || ^6.0.0"` while `node-sass@^7.0.3` is installed:
   - Upgrade `sass-loader@12.3.0` to latest. The major version change only marks the dropped support for nodejs < 14.

The peer dependency warning on `schema-utils@3.1.1` (peer dependency over `ajv@^6.9.1`) couldn't be solved because it doesn't look like a good idea to downgrade `ajv` and it's not possible to upgrade all packages requiring it as `raw-loader` has not been updated with a newer version. 

The preferred solution here might be to drop raw-loader and use the [asset-modules](https://webpack.js.org/guides/asset-modules/) provided by webpack 5 but that's another subject better left for some future PR.

Added an "engine" requirement to enforce the new min nodejs version of `>=14.15.0`, which is still generous given the official [maintenance timeline](https://nodejs.org/en/about/previous-releases). It is also in line with the requirement of #242. Still too recent for Ubuntu 18 though.

I checked the old/new `dist/` folders and it reported no change after the update.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8601 @@
 {
     "name": "voyager",
     "version": "0.35.0",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
     "requires": true,
+    "packages": {
+        "": {
+            "name": "voyager",
+            "version": "0.35.0",
+            "license": "Apache-2.0",
+            "dependencies": {
+                "ajv": "^8.6.2",
+                "buffer": "^6.0.3",
+                "client-zip": "^2.0.0",
+                "clone-deep": "^4.0.1",
+                "express": "^4.18.2",
+                "filenamify": "^4.3.0",
+                "fs-extra": "^10.0.0",
+                "handlebars": "^4.7.7",
+                "hotkeys-js": "^3.8.7",
+                "lit-element": "^2.5.1",
+                "lit-html": "^1.4.1",
+                "moment": "^2.29.4",
+                "morgan": "^1.10.0",
+                "nodemon": "^3.0.1",
+                "postcss": "^8.4.31",
+                "process": "^0.11.10",
+                "readable-stream": "^3.6.0",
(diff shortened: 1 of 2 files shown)


gjcope:
Looks good to me, thanks! Similar to #242 just looking to make sure it's not disruptive to non-developer users. I should be able to bundle these together when we are ready to move.

gjcope:
@sdumetz Would really like to make this jump but need the package to build with the included docker profile as well. I tried bumping to Ubuntu 20 and Node 16 but it freezes for me on build. I know you are not using it, but if you have a chance could you just see if https://github.com/Smithsonian/dpo-voyager/commits/dev-upgrade/ will build for you? It would be great to get somewhat back on track with Node version.

sdumetz:
Hi,

I checked-out the branch and the docker image builds fine (both development and production). If I delete `dist/` and `services/server/bin/`, it builds the client and server before starting the server

`npm install --dev` is redundant as it is the default behavior, but I didn't see it causing any problem otherwise.

Didn't test if the server was working properly, I stopped at `Server ready and listening on port 8000`.

I'm not familiar with those scripts though so I'm not even sure I ran them properly?

gjcope:
@sdumetz Thanks for taking a look. Good to know it is at least building for you. It must be something specific about my configuration. I added in the install --dev because after bumping the Node version it was only installing production dependencies when node_env == production. But maybe that's a symptom of whatever else is going on. Thanks again for taking a look.

gjcope:
Merged here: https://github.com/Smithsonian/dpo-voyager/tree/dev-upgrade - finally got a solid build. Thanks for the PR and sorry for the delay!

PR #245, opened 2024-01-16
use alpha mix for zonemap blend

sdumetz:
While testing the "overlayMap" features, I noticed the `USE_ZONEMAP` GLSL fragment ignored the texture's alpha.

It completely break usage of PNG/Webp textures with transparency.

I propose updating the blend to a simple alpha mix that solves the problem.

Of course I can visually achieve the same result by putting a black background on my texture so it might not be considered a bug. However it's not consistent with how textures with alpha are generally handled.

Another reason I'd want this is to be able to add customizez "blend modes" for overlayMap in the future. The current "additive blend" cannot work on very light-colored models and that's bound to be a problem in some cases.

### DIFF
--- a/source/client/shaders/uberPBRShader.frag
+++ b/source/client/shaders/uberPBRShader.frag
@@ -240,7 +240,7 @@ void main() {
 
 	#ifdef USE_ZONEMAP
 		vec4 zoneColor = texture2D(zoneMap, vZoneUv);
-		gl_FragColor += vec4(zoneColor.rgb, 1.0);
+		gl_FragColor += mix(gl_FragColor, vec4(zoneColor.rgb, 1.0), zoneColor.a);
 	#endif
 
 	#include <tonemapping_fragment>


gjcope:
Would

gjcope:
Would
`gl_FragColor += mix(vec4(0.0, 0.0, 0.0, 1.0), vec4(zoneColor.rgb, 1.0), zoneColor.a);`
work for your use cases?

My issue with it as-is is that since we are getting cumulative color, transparent areas of the .pngs will then get doubled, resulting in a lightening. We would like those areas to stay the same.

gjcope:
@sdumetz Do they above changes work for you?

sdumetz:
Hi,
Sorry I’m a bit overwhelmed right now.

Your solution is better.

gjcope:
No worries, I can make the update and tag you as contributor if it works better for you.

sdumetz:
I don’t care much for attribution details, especially for small changes like this. Do not hesitate to cherry-pick and modify my propositions as needed if that makes things simpler or faster for you!

Just mention the PR in your commit so I’ll know it has been handled.

gjcope:
Included in v0.38.0

PR #247, opened 2024-01-23
fix USE_CLEARCOAT in main shader fragment

sdumetz:
The current implementation of clearcoat's outgoingLight references variables that don't exist in three v0.157.

When using a model that triggers `USE_CLEARCOAT`, we get an error : 
```

FRAGMENT

ERROR: 0:3265: 'clearcoatSpecularDirect' : undeclared identifier
ERROR: 0:3265: 'clearcoatSpecularIndirect' : undeclared identifier
```

I rolled back the line to match the current  implementation.

Updating three to a later version (r158 and up) would also be a valid fix but I don't know if it will break anything else.

### DIFF
--- a/source/client/shaders/uberPBRShader.frag
+++ b/source/client/shaders/uberPBRShader.frag
@@ -226,7 +226,7 @@ void main() {
 
 		vec3 Fcc = F_Schlick( material.clearcoatF0, material.clearcoatF90, dotNVcc );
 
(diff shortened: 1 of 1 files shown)


gjcope:
There is also an issue with Sheen. We will update to r158 to fix both. (https://github.com/Smithsonian/dpo-voyager/commit/07af1459a604ec0de3b245f13af3e1df135d5acf)

gjcope:
Please check https://github.com/Smithsonian/dpo-voyager/tree/rc-37 for resolution

gjcope:
This has been fixed in v0.37.0 with a Three.js update as referenced above.

PR #249, opened 2024-01-26
register error in loadingManager when JSONWriter.put fails

sdumetz:
If for whatever reason `JSONWriter` fails to put the updated document, it causes loadingManager to be indefinitely busy.

This change properly registers the failure.


gjcope:
Merged to rc- 38 https://github.com/Smithsonian/dpo-voyager/commits/rc-38/

PR #250, opened 2024-02-05
minimal new lights integration

sdumetz:
This is a follow-up of #215. I removed the interface to add new lights, which will be the subject of another PR.

This is currently a work in progress, though we feel confident enough that it doesn't trigger regression to deploy it on our test service.

Work to be done :

- [ ] Improve Lights Helpers look and feel
- [ ] Usable config options for RectLight
- [ ] Recognizable icons for all light types (currently all using the same icon
- [ ] proper scaling when autoZoom is enabled (probably won't fix until the many quirks of autoZoom itself are taken care of and I feel confident reenabling it)

I don't know if/when I will be able to work on this but in any case I will do a follow-up and mark the PR as ready when I feel we have tested this enough to be production-ready from our point of view.

Pending this, I have this example of a good use-case for spot/point lights :

[Lascaux Cave](https://ecorpus.holusion.com/ui/scenes/Lascaux_1_composite/view)


gjcope:
@sdumetz Thanks for resubmitting. Also, would you mind opening an issue for "the many quirks of autoZoom itself"?

source/client/applications/coreTypes.ts
```diff
@@ -76,9 +76,21 @@ import NVDocuments from "../nodes/NVDocuments";
...
+import CVHemisphereLight from "client/components/CVHemisphereLight";
+import CVRectLight from "client/components/CVRectLight";
 
 ////////////////////////////////////////////////////////////////////////////////
 
+export const lightTypes = [
+    CVDirectionalLight,
+    CVPointLight,
+    CVSpotLight,
+    CVAmbienLight,
```
gjcope, line 89: Should be CVAmbientLight

sdumetz:
Updated to point on `ff-scene/master`.

Also added some per-light-type icons :
icons

moved everything to a new `lights/` folder to keep them together.

They are by no means final but I think we need to be able to identify different lights without opening the settings. It sqould be easy enough to update them later with a better design.

gjcope:
Currently not building for me due to CVScene light abstraction. See comment in submission.

sdumetz:
Yep, was a bad type cast. Don't know it built like this at first, but I did a clean build and saw the error.

That's fixed in [8a45970](https://github.com/Smithsonian/dpo-voyager/pull/250/commits/8a45970bdf0c38d8be25dc01eb0c3e10b0089dcb)

PR #253, opened 2024-02-19
Fix #252 set mediaManager.rootUrl before calling baseUrlValid

### LINKED ISSUE #252: A PROPFIND request is sent to the wrong address from voyager-story if root is provided
I'm using voyager-story from a `/ui/`route in my application, referencing a scene in a separate `/scenes/` folder.

It works fine excepts that a `PROPFIND /ui/scenes/:scene_name` gets sent before `CAssetManager.rootUrl` could be properly assigned in [StoryApplication.evaluateProps()](https://github.com/Smithsonian/dpo-voyager/blob/master/source/client/applications/StoryApplication.ts#L193).

In the end it boils down to `CVMediaManager.baseUrl` not being the same thing as `CAssetManager.rootUrl` until re-set in `StoryApplication.evaluateProps()` for reasons I do not really understand.

If that's a desired behavior and the reassignment can't be made before the request is sent, I could write a patch where `WebDAVProvider` makes use of  an `AbortController` to cancel the request in-flight, which would at least hide the error.

sdumetz:
the previous order of operations would cause a request to be fired to the wrong URL if `root` is set to something other than `/<scene_name>`(eg. a `scenes` subfolder).

### DIFF
--- a/source/client/applications/StoryApplication.ts
+++ b/source/client/applications/StoryApplication.ts
@@ -187,10 +187,10 @@ export default class StoryApplication
         }     
 
         // if dragging/dropping have to assume that a non-loading url is still valid
+        this.mediaManager.rootUrl = this.assetManager.baseUrl;
         if(props.dragdrop === true) {
             this.assetManager.ins.baseUrlValid.setValue(true);
         }
-        this.mediaManager.rootUrl = this.assetManager.baseUrl;
 
         const tasks = this.system.getMainComponent(CVTaskProvider);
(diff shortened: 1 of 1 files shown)


gjcope:
Looks like previously this was causing a WebDAV call in Standalone mode (https://github.com/Smithsonian/dpo-voyager/commit/3f0d832a37bdb28c72dabf27885a8f99661fd732) but I'm not seeing it now. Can you confirm? Will merge if so. Thanks.

PR #274, opened 2024-05-09
Improve docker-compose installation method:

umer936:
1) proper variable usage in docker-compose.yml
2) put provision.sh inline of the Dockerfile
3) make container not rely on root user -- fixes npm permissions issue 
4) define all versions at top of Dockerfile
5) Bail early if /app not mounted

Feel free to modify as desired, but this means the container does not rely on npm being installed on the host system and it fixed the issues I was facing regarding npm failing permissions due to being run as root.


umer936:
hmm may need a note in the install instructions that the dpo-voyager folder needs to be writable by all (`chmod a+w .`) bc npm install makes `node_modules` and `package-lock.json` 

alternatively could have the folders (e.g. files, node_modules, and dist) included in the repo with just a .gitkeep file in the folders

gjcope:
Thanks for the PR @umer936, we will take a look.

gjcope:
> hmm may need a note in the install instructions that the dpo-voyager folder needs to be writable by all (`chmod a+w .`) bc npm install makes `node_modules`, `dist`, and `package-lock.json`

Can you describe the install process you are following in more detail? None of the installs we have on Windows or RHEL have a project folder that is writable by all.

umer936:
So the issue I was trying to solve with this is that I don't have NPM installed on my host system, just in the docker container and it was giving a lot of permissions issues. 

I had this happen on 2 machines by two of us, one on Windows and one of us on RockyLinux.

When I attempted to just run `docker-compose up`, I encountered errors like this (not using sudo): 
MicrosoftTeams-image (4)

and Google led me to that being due to the user in the Dockerfile being root and then npm being upset that I'm trying to run it as root. I figured this was due to conflicts in the user being used to build the container, hence the rework of the `Dockerfile`/`start.sh`/`provision.sh`.


That said, I started from scratch just now and didn't have an issue at all... I'm not sure where the permissions issue was stemming from. I haven't run through the commits you've done since last I pulled. Maybe it was an npm package version issue, or maybe it was user error. 

I'll close this issue/PR because I can't replicate it today. 

Let me know if you want me to commit the integrating `start.sh` into the `Dockerfile`, but also feel free to ignore.

gjcope:
@umer936 Thanks for the extra context. I'm definitely interested in this PR since if you encountered the issue twice, odds are that others have as well. I just want to understand the underlying issue before making the change. If you come across any more info on  what might have been different between your working/not working deployments, please do reopen. Thanks!

umer936:
I undid this commit on the branch I was working on on the Debian laptop and the issue returned. Windows is not having an issue. Will investigate again in a month or two.

gjcope:
Jumping back into this as we have encountered somewhat similar issues on a new vm with a current version of Docker. Everything worked if the Voyager directory owner was set to root and no other changes. Not a viable solution, but just curious if this works in your case as well. The updates in this commit, by themselves, didn't fix things in our case.

gjcope:
This should be fixed by #323 . Feel free to reopen with explanation if not. Thanks!

PR #280, opened 2024-06-11
Better other default language support

sdumetz:
I did a bunch of work trying to bring the multi-language support to a point where non-english speakers and particularly non-primarily-english scenes have a smooth experience.

**There was a number of issues:**

 - In a number of places, `ELanguageType.EN` was used interchangeably with `DEFAULT_LANGUAGE`. The fix is straightforward. `ELanguageType.EN` now used as a default only where it makes sense.
 - I removed every individual task's `ins.language` property. They were the source of a number or race conditions, while barely used in the code and I found it easier to remove them and use `CVLanguageManager`as the unique source of language state everywhere. This should also help avoid regressions in the future.

**Some backward-compatible changes : **

 - Fixing a race condition where the scene's language would get set by `ExplorerApplication` using a queryString then changed back to the document's defaults when the document loads. See [9bacad87](https://github.com/Holusion/dpo-voyager/commit/9bacad873446d401c6222079bb1bcae3b90b274c?diff=unified&w=0) and set the page's queryString to some other language to trigger the race condition. I had to create a new `UNSET` language value. This has minimal impact on the system because it always get set before it really gets used.

**A breaking change that I feel was necessary: **

Added a `document.setups[0].language.languages` property describing the document's allowed locales. Without it I couldn't get the `activeLanguages` feature to work reliably in every case. Additionally it allows an user to manually choose which languages will be available in the scene. As a nice bonus, the language picker is auto-synchronized to allow only those options everywhere (Tasks, CollectionPanel, LanguageMenu).

It won't break older scenes, gathering the allowed languages from existing data.

**Note :** it requires https://github.com/Smithsonian/ff-graph/pull/1 to work properly.

I left every commit separate to make those changes more readable: each message should be relatively self-describing. Most of them could be squashed together once this is ready to merge.

We tested it on a lot of our scenes already  (French-only, English-only, French-first, English-first) so this shouldn't cause too much unexpected trouble.

### DIFF
--- a/source/client/components/CVLanguageManager.ts
+++ b/source/client/components/CVLanguageManager.ts
@@ -46,11 +46,12 @@ export default class CVLanguageManager extends Component
 
     protected static readonly ins = {
         enabled: types.Boolean("Language.Enabled", false),
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN),
+        language: types.Enum("Interface.Language", ELanguageType, ELanguageType[DEFAULT_LANGUAGE]),
     };
 
     protected static readonly outs = {
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN),
+        /* exception to default language: in absence of any dictionary, this is always EN */
+        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN ),
     };
 
     ins = this.addInputs(CVLanguageManager.ins);
@@ -114,7 +115,7 @@ export default class CVLanguageManager extends Component
         const language = ELanguageType[data.language || "EN"];
 
         if(language != outs.language.value && ins.language.value === outs.language.value) {
-            ins.language.setValue(isFinite(language) ? language : ELanguageType.EN);
+            ins.language.setValue(isFinite(language) ? language : ELanguageType[DEFAULT_LANGUAGE]);
         }
     }
 
--- a/source/client/components/CVLanguageManager.ts
+++ b/source/client/components/CVLanguageManager.ts
@@ -75,7 +75,11 @@ export default class CVLanguageManager extends Component
         return this._activeLanguages;
     }
 
-    nameString()
+    /**
+     * 
+     * @returns Full text string of the currently selected language
+     */
+    nameString() :string
     {
         return ELanguageStringType[ELanguageType[this.ins.language.value]];
     }
--- a/source/client/components/CVLanguageManager.ts
+++ b/source/client/components/CVLanguageManager.ts
@@ -40,7 +40,7 @@ export default class CVLanguageManager extends Component
     static readonly text: string = "Language";
     static readonly icon: string = "";
 
-    private _activeLanguages: ILanguageOption[] = [];
+    private _activeLanguages: {[key in TLanguageType]?: ILanguageOption} = {};
     private _translations: ITranslation = {};
 
     static readonly isSystemSingleton = true;
@@ -72,7 +72,7 @@ export default class CVLanguageManager extends Component
         return this.getMainComponent(CVAssetReader);
     }
     get activeLanguages() {
-        return this._activeLanguages;
+        return Object.values(this._activeLanguages);
     }
 
     /**
@@ -98,13 +98,14 @@ export default class CVLanguageManager extends Component
     {
         const { ins, outs } = this;
         
-        if(this.activeLanguages.length == 0) {
+        if(this.activeLanguages.length == 0 && ins.language.value == outs.language.value) {
             this.addLanguage(outs.language.value);
             //return;
         }
         
         if (ins.language.changed && ins.language.value != outs.language.value) {
             const newLanguage = ins.language.value;
+            this.addLanguage(newLanguage);
             this.assetReader.getSystemJSON("language/string.resources." + ELanguageType[this.ins.language.value].toLowerCase() + ".json").then( json => {
                 this._translations = json;
                 this.updateLanguage(newLanguage);
@@ -121,10 +122,11 @@ export default class CVLanguageManager extends Component
         const { ins, outs } = this;
         data = data || {} as ILanguage;
 
-        const language = ELanguageType[data.language || "EN"];
+        const language = ELanguageType[data.language || "EN"] ?? ELanguageType[DEFAULT_LANGUAGE];
 
-        if(language != outs.language.value && ins.language.value === outs.language.value) {
-            ins.language.setValue(isFinite(language) ? language : ELanguageType[DEFAULT_LANGUAGE]);
+        //If language has already been set, don't overwrite it.
+        if(ins.language.value < 0) {
+            ins.language.setValue(language);
         }
     }
 
@@ -138,11 +140,7 @@ export default class CVLanguageManager extends Component
     }
 
     addLanguage(language: ELanguageType) {
-        const exists = this._activeLanguages.find(element => element.id === language)
-
-        if(!exists) {
-            this._activeLanguages.push({ id: language, name: ELanguageStringType[ELanguageType[language]] });
-        }
+        this._activeLanguages[ELanguageType[language]] ??= { id: language, name: ELanguageStringType[ELanguageType[language]] };
     }
 
     getLocalizedString(text: string): string
--- a/source/client/components/CVLanguageManager.ts
+++ b/source/client/components/CVLanguageManager.ts
@@ -20,6 +20,7 @@ import { ILanguage, ILanguageOption } from "client/schema/setup";
 import { ELanguageType, TLanguageType, ELanguageStringType, DEFAULT_LANGUAGE } from "client/schema/common";
 import CVAssetReader from "./CVAssetReader";
 import { ITagUpdateEvent } from "./CVModel2";
+import { enumToArray } from "@ff/core/types";
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -46,7 +47,11 @@ export default class CVLanguageManager extends Component
 
     protected static readonly ins = {
         enabled: types.Boolean("Language.Enabled", false),
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType[DEFAULT_LANGUAGE]),
+        language: types.Enum("Interface.Language", ELanguageType, {
+            preset: ELanguageType[DEFAULT_LANGUAGE],
+            enum: ELanguageType,
+            options: enumToArray(ELanguageStringType).map(key => ELanguageStringType[key])
+        }),
     };
 
     protected static readonly outs = {
--- a/source/client/ui/PropertyOptions.ts
+++ b/source/client/ui/PropertyOptions.ts
@@ -14,6 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+	
+import {live} from "lit-html/directives/live";
 
 import Property from "@ff/graph/Property";
 import CustomElement, { customElement, property, PropertyValues, html } from "@ff/ui/CustomElement";
@@ -64,6 +66,7 @@ export default class PropertyOptions extends CustomElement
             const property = changedProperties.get("property") as Property;
             if (property) {
                 property.off("value", this.onUpdate, this);
+                property.off("change", this.onUpdate, this);
             }
             if (this.property) {
                 this.property.on("value", this.onUpdate, this);
@@ -102,8 +105,7 @@ export default class PropertyOptions extends CustomElement
 
         return html`
             <label class="ff-label ff-off">${name}</label>
-            <select ?multiple=${property.isMulti()} class="sv-property-field" @change=${(e)=>{
-                console.debug("Select value : ", e.target.value);
+            <select ?multiple=${property.isMulti()} .value=${live(value)} class="sv-property-field" @change=${(e)=>{
                 this.property.setValue(e.target.value)
             }}>
                 ${optionsList}
--- a/source/client/ui/PropertyOptions.ts
+++ b/source/client/ui/PropertyOptions.ts
@@ -67,6 +67,7 @@ export default class PropertyOptions extends CustomElement
             }
             if (this.property) {
                 this.property.on("value", this.onUpdate, this);
+                this.property.on("change", this.onUpdate, this);
             }
         }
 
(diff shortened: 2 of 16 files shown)


source/client/components/CVLanguageManager.ts
```diff
@@ -46,11 +46,12 @@ export default class CVLanguageManager extends Component
...
     protected static readonly ins = {
         enabled: types.Boolean("Language.Enabled", false),
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN),
+        language: types.Enum("Interface.Language", ELanguageType, ELanguageType[DEFAULT_LANGUAGE]),
     };
 
     protected static readonly outs = {
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN),
+        /* exception to default language: in absence of any dictionary, this is always EN */
+        language: types.Enum("Interface.Language", ELanguageType, ELanguageType.EN ),
```
gjcope, line 67: Why an exception? In the absence of any dictionary, won't DEFAULT_LANGUAGE be EN anyways?

gjcope (commented):
Can you be more specific about the issue this is addressing? Property views in the UI should already be showing the full text except for the Explorer button that opens the language menu.

source/client/components/CVLanguageManager.ts
```diff
@@ -46,7 +47,11 @@ export default class CVLanguageManager extends Component
 
     protected static readonly ins = {
         enabled: types.Boolean("Language.Enabled", false),
-        language: types.Enum("Interface.Language", ELanguageType, ELanguageType[DEFAULT_LANGUAGE]),
+        language: types.Enum("Interface.Language", ELanguageType, {
```
gjcope, line 49: Can you be more specific about the issue this is addressing? Property views in the UI should already be showing the full text except for the Explorer button that opens the language menu.

source/client/components/CVLanguageManager.ts
sdumetz: In the absence of a dictionary, the "output" language is always `EN` (or rather "untranslated UI strings", in English by convention).

This is not the same as `DEFAULT_LANGUAGE`, as someone might set it to another value.

Because of this I left this comment to not be tempted to replace `ELanguageType.EN` with `DEFAULT_LANGUAGE` as it was done everywhere else because this would be wrong and introduce again places where English is confused with DEFAULT_LANGUAGE, which defeats the idea of a DEFAULT_LANGUAGE global.

source/client/components/CVLanguageManager.ts
gjcope: Never mind, I see this supports the other task changes.

sdumetz: Those properties were using a mirror of this property to provide the full text. This required a lot `synchLanguage()` that were prone to misuse. Seeing that removing all those properties ([49cf2a4](https://github.com/Smithsonian/dpo-voyager/pull/280/commits/49cf2a4b4ec0903430a7fef353e14d3f55381754)..[a0dd844](https://github.com/Smithsonian/dpo-voyager/pull/280/commits/a0dd84469327779ddea8f73f367f709474036010)) was really simple I thought it more elegant to always rely on CVLanguageManager's property. However the selectors then displayed to language's codeString ("FR", "EN"), which was a regression that this fixes preventively.

source/client/ui/PropertyOptions.ts
```diff
@@ -120,22 +124,47 @@ export default class PropertyOptions extends CustomElement
         let buttons;
         if (indexMap) {
             buttons = indexMap.map(index =>
-                html`<ff-button role="radio" aria-checked=${index === value ? true : false} tabbingIndex=${indexMap.includes(value) ? (index === value ? 0 : -1) : (index === indexMap[0] ? 0 : -1)} .text=${language ? language.getLocalizedString(options[index]) : options[index]} .index=${index} .selectedIndex=${value} @click=${this.onButtonClick}>
```
gjcope, line 123: I'll have to test but I believe we still need aria-checked for screen reader support.

source/client/components/CVLanguageManager.ts
gjcope: Fair point, thanks for the explanation.

source/client/components/CVLanguageManager.ts
```diff
@@ -135,25 +152,37 @@ export default class CVLanguageManager extends Component
 
         return {
             language: ELanguageType[ins.language.getValidatedValue()] as TLanguageType,
+            languages: ins.languages.value.map(l=>ELanguageType[l]) as TLanguageType[],
         };
     }
 
+    /**
+     * Reference a list of languages used in the document
```
gjcope, line 160: Not sure I'm understanding the value of this. Can you provide a specific use case?

source/client/ui/explorer/LanguageMenu.ts
```diff
@@ -84,7 +84,7 @@ export default class LanguageMenu extends Popup
             </div>
             <div class="ff-flex-row">
                 <div class="ff-scroll-y sv-scroll-offset" role="listbox">
-                    ${language.activeLanguages.map((language, index) => this.renderEntry(language, index))}
+                    <sv-property-options class="ff-flex-column" buttons .property=${language.ins.language}></sv-property-options>
```
gjcope, line 87: I may be missing something from another commit, but this doesn't seem correct to me. language.ins.language should be all supported languages (i.e. all languages we have dictionaries for) and the Explorer language menu should have all languages actively used in the scene.

sdumetz: I opted to split the "list of accepted languages" (a subset of "all dictionaries") into `language.ins.languages`, whose value is then used as possible options for `language.ins.language` ; then always propose just this subset of languages until another language is enabled in the scene.

The reason behind this is something I saw a lot of users do during scene editing :
- edit some annotation/article in a french-only scene
- click on "Français", switch to "English" (misclick, or looking for something else, or not knowing this scene is not translated)
- Save the scene at any point in the future
- The scene is now FR/EN forever, even though most/all EN values are undefined. There is no way to go back to "french-only"

To prevent this I rather liked the concept of this intermediate list of languages that makes it unequivocaly clear what languages are expected to be supported in this scene.

We might go the other route and not have this list. The list of available languages would then be "any language for which something is defined" but then : 

1. we can't easily delete a language for a scene once it has been touched
2. we have to thoroughly search through the scene for all translation dictionaries
3. inside the Story editor we will have all available languages to select from. Keeping in mind that I might be required to commit to some basic localization for a large chunk of languages spoken in the EU in the near future, that's a quite a list to have.

source/client/ui/PropertyOptions.ts
sdumetz: Fixed in 135829e (will squash it together once all discussions are resolved)

source/client/ui/explorer/LanguageMenu.ts
gjcope: Thanks for the explanation. I understand the motivation, and acknowledge the issue, but am not sure I agree this is the best route forward.

> * edit some annotation/article in a french-only scene
> * click on "Français", switch to "English" (misclick, or looking for something else, or not knowing this scene is not translated)
> * Save the scene at any point in the future
> * The scene is now FR/EN forever, even though most/all EN values are undefined. There is no way to go back to "french-only"

I think this is only true with articles, as switching languages currently creates a new html article asset. With annotations and titles, if the data is undefined on save it shouldn't be getting written to the file. If there are cases where it is, that sound like a bug. 

> 
> To prevent this I rather liked the concept of this intermediate list of languages that makes it unequivocaly clear what languages are expected to be supported in this scene.
> 

How do you see this list being defined? From my current perspective, this isn't something we could expect from our users.

> We might go the other route and not have this list. The list of available languages would then be "any language for which something is defined" but then :
> 
> 1. we can't easily delete a language for a scene once it has been touched
> 2. we have to thoroughly search through the scene for all translation dictionaries
> 3. inside the Story editor we will have all available languages to select from. Keeping in mind that I might be required to commit to some basic localization for a large chunk of languages spoken in the EU in the near future, that's a quite a list to have.

For 3. this is intentional. I don't think it's any different from standard localization lists in many sites/apps.

I'm going to think a little bit more on alternatives to address this issue.

gjcope:
Still considering the "active language" commit, but other than that this PR looks great. Thanks for submitting and I appreciate the commits being broken up!

sdumetz:
Clearly an area where there is more than one "good" solution. I felt I had to propose a solution as a starting point but I am ready to consider other solutions if they fit the greater picture better.

For the selection of a list of "scene languages", our users generally have a very clear idea of what this list should be, hence the proposal. 

If you have something in mind please let me know. Otherwise I'll try to have another look at it shortly, taking your comments into account to see if I can come up with a better solution.

gjcope:
I think my question re: the list is more about how that is operationalized. Would you as an admin be setting this list before they start in each scene file? Would the end user be setting this list? It isn't clear to me how it would be configured, even if you know what this list would be.

gjcope:
Talked to some more folks about this, but still really interested in how you see this being configured. But something else to consider:

- Language list is still driven by the full list of supported languages. If the list gets too long, there are some UI improvements we could add that are pretty common to long localization selections.
- Languages are still added to the 'active' list by selecting in a language dropdown, but adding in a pop-up to give users an out if it was unintended. (i.e. "French is not currently used in your scene. Would you like to add it?")
- Providing some sort of UI for removing languages from the supported list which would also remove any associated data from the scene. I'm less clear on how this piece would fit in.

The advantage I see here is not relying on a static list and having the 'active' list be more data driven. But again, looking for your config plan as maybe you were thinking something similar.

gjcope:
> In our case, the end user is generally the one creating the scene. In the current state it is a list of buttons to choose from in the "Collection" tab, which isn't even greatly discoverable.
> 

I see. I think the alternate proposal builds it into the existing workflow so may be a little more user friendly.

> It would also be possible to call `addLanguage()` from the initializers of `Annotation`, `Article`... but this would add back language management code all over the place, wouldn't it?

We already have this code for annotations and title (e.g. https://github.com/Smithsonian/dpo-voyager/blob/c08cb20741d5f0aaadd0aa8a5229152e30d4e017/source/client/components/CVMeta.ts#L79). It doesn't look like your PR touched it. But it also looks inconsistent for articles, so that would need to be added. I don't think this is "all over the place", just the three components with localized content. If a new content component was added, it would need to be implemented for that too, but that seems reasonable. 

I'm not sure how we would centralize this anymore without reimplementing the language enum to encapsulate it but that seems extreme.

> 
> To handle the potentially growing list of available languages, one idea would be to be able to group options, presenting the active languages first and the other second. However as I've been thinking about this, I now think that's better left for a future time because :
> 
I was thinking the same thing about pushing the active languages to the top, but agree it is not a priority at the moment.
> 
> I'm curious about your opinion on whether `setup.language.languages` is desirable or not, then I think I'll have everything I need to improve [d58414e](https://github.com/Smithsonian/dpo-voyager/pull/280/commits/d58414e2f3354cc1899bcd226e4914966e3193bf)..[26a4574](https://github.com/Smithsonian/dpo-voyager/pull/280/commits/26a4574fa6e07b3f191cd0c671b0aa4b6c083efd)

I don't have a problem with it and see how it could serve as a shortcut to loading active languages from established scenes, but I don't know that it is required if we go with the above changes. My remain issue is how a user would remove things from the active list.

sdumetz:
I force-pushed to collapse the commits from [d58414e](https://github.com/Smithsonian/dpo-voyager/pull/280/commits/d58414e2f3354cc1899bcd226e4914966e3193bf) onwards because all the reverts started getting messy.

Did a few more tests. Language is set order of precedence :
1. to English on page load (forced by the string translation system until a dict is available)
2. to DEFAULT_LANGUAGE (no-op if `DEFAULT_LANGUAGE = "EN"`)
3.  to `setup[0].language.language` on document load
4. to the `lang` queryString value. If it doesn't map to a supported language in the scene, it is forcefully added to the list of active languages.
5. to the user-selected value from any of the `<sv-property-options>` selectors (every language available) or `LanguageMenu`(restricted to `CVLanguageManager.activeLanguages`


Don't know for sure if **4.** is the most desirable behavior? Checking `activeLanguages` before switching to the value of `?lang` seems not practical because it would introduce back a race condition between this and the document load. On the  other hand not adding it to `activeLanguages` would make the user unable to switch to another language than back to it.

Should be OK to merge now, please let me know if there is any remaining issue that I overlooked or if you uncover any bug this introduces.

gjcope:
Did some quick testing.

One bug I noticed can be reproduced by:

1. Change to new language
2. Add new title
3. Select annotation (with Collection tab still visible)
4. Change language under annotation
5. Change back to language from step 1
6. Collection tab language will now say 'English' regardless of selected language. I assume it is because the index is null for some reason but haven't been able to look into it yet.


> 4. to the `lang` queryString value. If it doesn't map to a supported language in the scene, it is forcefully added to the list of active languages.

'lang' is currently a property that really only applies to explorer, so if it doesn't map to a supported language that would mean there is no content for that language and I would think it should be ignored.

Other than those two things, the PR looks good it just doesn't seem like we are addressing the original issue (unless I'm missing something). Users can still unintentionally add languages with no way to remove them, right? Seems like we still need something like a confirmation pop-up when adding new languages and some way to remove them. But maybe this is for another day. There is still a lot of valuable cleanup in this PR.

sdumetz:
> I assume it is because the index is null for some reason but haven't been able to look into it yet

Looks like a bad use of litElement templates. I can reproduce the issue, the HTML is properly updated (in the inspector), but the wrong value is shown. It looks like a [known issue](https://stackoverflow.com/questions/76231648/lit-html-still-use-the-previous-dropdown-selected-index-when-rendering-a-dropdow) and using the `live()` directive seems to fix it (3cbbeb57).


> 'lang' is currently a property that really only applies to explorer, so if it doesn't map to a supported language that would mean there is no content for that language and I would think it should be ignored.

This is reasonable, however I don't know if we can without introducing back some race conditions between this and the document deserializer? Currently we don't know what will be the list of available language from where the queryString is parsed.


Regarding the issue, that's me going all over the place. The core issue would be "there is currently no way to make Voyager reliably stick to a language other than English". It's just that I know I18n when done wrong tends to come back at you and bite hard so I try to think ahead.

I'm Ok with the other issues left for later. For english-first scenes, this PR is effectively just a cleanup job but for scenes that defaults to another language it is already a great improvement.

gjcope:
Did you push 3cbbeb57? I'm not seeing it.

> This is reasonable, however I don't know if we can without introducing back some race conditions between this and the document deserializer? Currently we don't know what will be the list of available language from where the queryString is parsed.

Good point. It's not pretty, but maybe https://github.com/Smithsonian/dpo-voyager/blob/c08cb20741d5f0aaadd0aa8a5229152e30d4e017/source/client/applications/ExplorerApplication.ts#L692 is changed to a new 'initLanguage' function or secondary property or something like that to separate setting language via api/attribute so it can be evaluated when available.

sdumetz:
Sorry, forgot to push...

If that's OK with you, I'd wait to see if the problem actually happens on a regular basis in the real world before implementing a solution. I don't know if it would actually happen to have a queryString not matching the referenced scene.

If it only happen once in a while, I vote to leave it as it is : it's the simplest solution. And if it becomes a problem we can fix it later using your idea or something else that might come up depending on the context.

gjcope:
Sounds good. Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-42

PR #281, opened 2024-06-12
replace node-sass with sass to support build over node22

sdumetz:
[node-sass](https://www.npmjs.com/package/node-sass) (or rather its underlying library, libsass) has been [deprecated](https://sass-lang.com/blog/libsass-is-deprecated/) in favor of dart-sass (the [sass package](https://www.npmjs.com/package/sass).

Consequently, node-sass no longer provides new versions with prebuilt binaries for newer nodejs versions, causing the install to be super-long on node-v18.x (only supported by `node-sass>=8.0`). I couldn't get it to build for node-v22.x (current stable release).

With the dependency swap, voyager builds perfectly on every node version from 16.x to 22.x (didn't test anything earlier)

Using a clean install before/after this patch with  `node-v14.x` I compared those files:

- `dist/js/voyager-explorer.min.js`  : different in production (unreadable output), but in development the only different line was the version header (due to the commit not matching a git tag)
- `dist/css/` : strictly equal in production


sdumetz:
Force-pushed [de80df3](https://github.com/Smithsonian/dpo-voyager/pull/281/commits/de80df3d6f915d1cd9b40f7128aed24fe87204d8) over [1234831](https://github.com/Smithsonian/dpo-voyager/pull/281/commits/1234831c6cebe0d2b902986d6403c9eeaa975ec5) to check-in a package-lock.json without the optional dependencies.

In [853abf3](https://github.com/Smithsonian/dpo-voyager/pull/281/commits/853abf37ceb9427e5f96657dafafb9675168e390) I updated the CSS to address a [deprecation warning](https://sass-lang.com/documentation/breaking-changes/slash-div/) from Dart-SASS.

The change is quite trivial and backward compatible.

gjcope:
Thanks for the PR! Should be good as we have pushed our build image to 16.x but I will give it a test.

gjcope:
Merged with https://github.com/Smithsonian/dpo-voyager/tree/rc-43

PR #285, opened 2024-06-27
Keep Copyright notices

sdumetz:
The document's `asset.copyright` field is currently always erased on save : the value is hardcoded in `CVDocument`.

I'd like it to be kept the same through saves as we have multiple sources that have different copyright requirements. A scene modification shouldn't change this field.

Additionally, I'd like to be able to edit the notice in the UI so I propose to add a text field in the collection panel. But maybe that's just me and this would be undesirable for everyone else? I'm OK handling this outside of voyager if necessary as long as it don't get replaced after each scene change.

### DIFF
--- a/source/client/components/CVDocument.ts
+++ b/source/client/components/CVDocument.ts
@@ -66,6 +66,7 @@ export default class CVDocument extends CRenderGraph
         download: types.Event("Document.Download"),
         title: types.String("Document.Title"),
         intro: types.String("Document.Intro", ""),
+        copyright: types.String("Document.Copyright", "(c) Smithsonian Institution. All rights reserved."),
     };
 
     protected static readonly outs = {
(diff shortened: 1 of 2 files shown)


gjcope:
This is great, and an important point for supporting collaborators.

Only thing I notice is that maybe the text box should not be indented since it is not active-language dependent.

sdumetz:
Oh, right. I amended the commit.

gjcope:
Thanks! Merged with https://github.com/Smithsonian/dpo-voyager/tree/rc-43

Note - I think it is a little out of place in the Collection tab, since copyright is not stored in the collection element of the scene file, but we don't really have a better place to put it at the moment. Maybe we consider another scene-level element if more of this sort of data comes up...

sdumetz:
> Note - I think it is a little out of place in the Collection tab, since copyright is not stored in the collection element of the scene file, but we don't really have a better place to put it at the moment. Maybe we consider another scene-level element if more of this sort of data comes up...

Yes. I saw nowhere else to put this and didn't want to add another panel. Collection seemed close enough for now. It should definitely be moved if something better comes up.

PR #296, opened 2024-07-26
add back ff-color-edit popup. Fix #257

### LINKED ISSUE #257: Use the color wheel to select an annotation color

sdumetz:
This PR reintroduces `<ff-color-edit>` instead of the native `<input type="color">` to solve accessibility issues.

I more or less copied-back the original implementation that existed before https://github.com/Smithsonian/dpo-voyager/pull/259 with a few changes:

- some styles updates (minor)
- use a `pickerActive` reactive property instead of forcing requestUpdate (more concise)
- Register `document.addEventListener("click",/* */)` to close picker when clicking somewhere else only when picker is active (small performance optimization)
- Add shortcut `Enter` to close picker

Sorry for the long delay in fixing this and more generally for the inconvenience caused by https://github.com/Smithsonian/dpo-voyager/pull/259, I clearly didn't test it enough before submitting.

### DIFF
--- a/source/client/ui/PropertyColor.ts
+++ b/source/client/ui/PropertyColor.ts
@@ -21,12 +21,10 @@ import Property from "@ff/graph/Property";
 import CustomElement, { customElement, property, PropertyValues, html } from "@ff/ui/CustomElement";
 
 import "@ff/ui/Button";
-import { IButtonClickEvent } from "@ff/ui/Button";
-
 import "@ff/ui/ColorEdit";
-import { IColorEditChangeEvent } from "@ff/ui/ColorEdit";
 
-import {getFocusableElements, focusTrap} from "../utils/focusHelpers";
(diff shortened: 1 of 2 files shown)


gjcope:
No worries and thanks, looks good. Merged with [rc-44](https://github.com/Smithsonian/dpo-voyager/tree/rc-44)

PR #302, opened 2024-09-04
accept other audio formats for AudioTaskView

sdumetz:
AudioTaskView currently only accepts `mp3` audio uploads. While this is the only compressed audio format that has 100% support (source: [caniuse.com](https://caniuse.com/?search=audio%20format) across browsers, some tools (eg: apple's audio recorder) now defaults to more modern formats.

I propose a softer requirement where partially-supported formats would just trigger an alert though I'll understand if you want to keep stricter requirements.

### DIFF
--- a/source/client/ui/story/AudioTaskView.ts
+++ b/source/client/ui/story/AudioTaskView.ts
@@ -157,10 +157,19 @@ export default class AudioTaskView extends TaskView<CVAudioTask>
         }
 
         const id = element.parentElement.parentElement.id;
-        const fileProp = id == "filename" ? this.task.ins.filepath : this.task.ins.captionPath;
-        const extText = id == "filename" ? ".mp3" : ".vtt";
+        const type = (id == "filename")? "audio": "subs";
+        const fileProp = (type == "audio") ? this.task.ins.filepath : this.task.ins.captionPath;
 
-        if(filename.toLowerCase().endsWith(extText)) {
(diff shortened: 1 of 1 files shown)


source/client/ui/story/AudioTaskView.ts
```diff
@@ -157,10 +157,19 @@ export default class AudioTaskView extends TaskView<CVAudioTask>
...
-        const fileProp = id == "filename" ? this.task.ins.filepath : this.task.ins.captionPath;
-        const extText = id == "filename" ? ".mp3" : ".vtt";
-
-        if(filename.toLowerCase().endsWith(extText)) {
+        const type = (id == "filename")? "audio": "subs";
+        const fileProp = (type == "audio") ? this.task.ins.filepath : this.task.ins.captionPath;
+
+        const ext = filename.toLowerCase().split(".").pop();
+        if(type === "subs" && ext != "vtt"){
+            Notification.show(`Unable to load - Only ${type === "subs"?".vtt":".mp3,.m4a"} files are currently supported.`, "warning");
```
gjcope, line 165: With this restructuring, the `type === "subs"` check isn't really necessary anymore

gjcope:
Yes, the support aspect is why we limited it originally. I will talk to my team and see about opening it up. I generally don't like the idea of supporting content that can ultimately make it look like things are broken, but I understand the need as things have moved beyond mp3.

PR #304, opened 2024-09-06
allow pasting color hex codes in sv-property-color. Fix #303

### LINKED ISSUE #303: Color widgets missing manual input
Recent color widget updates have improved consistency and accessibility, but we've now lost the ability to manually input a specific color. This is important when you are trying establish a standard look/feel across Voyager experiences. @sdumetz

sdumetz:
First commit implements the text input. It looks like this : 

image

Then I wanted to test the various input formats. For reference, `@ff/core/Color` is able to parse  3-6 chars hex, `rgb()`, `hsl()` notations. All of which are parsed OK, though they get auto-converted to 6-chars hex immediately. Dealing with this would probably require storing some internal string-state inside the `Color` class but would still be lost on save. Let me know if you think it's needed.

Then I wanted to test rgba notations. I noticed no property currently use rgba. There is `Material.BaseColor` from **CVModel2** that is mapped to a Vector4, but the error is the other way around: It gets assigned to a `THREE.Color` that have no alpha. I fixed that in 7073d361c9b16598f4aed8ccf4f2db29aaf425a7.

I still temporarily switched a property to RGBA to test things. It was broken in various minor ways, I fixed that in b8c886c6b975d5c729f1de9f4b2d7b472661aa8b, we might need it in the future.

Then I noticed the serialized alpha colors would be using the `rgba()`notation which overflows from our inputs. Since `#rrggbbaa` is now [widely supported](https://caniuse.com/css-rrggbbaa), I propose to change the default and add the ability to parse it in d9d5ba748a94309f11da1c3244d41e880c02d958. I modified the serialization technique because the bit shift doesn't like 32bits offsets. I also added unit tests even though they don't run as they are now, expect a PR about this soon.

02d11c6d49f43266d5d08d440f62fd3db252b529 adds a simple validity check for colors that fail to parse. It's invisible now but it will show if/when we add some style for invalid inputs.

### DIFF
--- a/libs/ff-core/source/Color.ts
+++ b/libs/ff-core/source/Color.ts
@@ -299,26 +299,20 @@ export default class Color implements IVector4
     {
         color = color.trim().toLowerCase();
         color = Color.presets[color] || color;
-
-        let result = color.match(/^#?([0-9a-f]{3})$/i);
-        if (result) {
-            const text = result[1];
+        let text = /^#?([0-9a-f]{3,8})$/i.exec(color)?.[1];
+        if (text && (text.length === 3 || text.length == 4)) {
             const factor = 1 / 15;
             this.x = Number.parseInt(text.charAt(0), 16) * factor;
             this.y = Number.parseInt(text.charAt(1), 16) * factor;
             this.z = Number.parseInt(text.charAt(2), 16) * factor;
-            this.w = alpha;
+            this.w = (text.length == 4)? Number.parseInt(text.charAt(3), 16) * factor:alpha;
             return this;
-        }
-
-        result = color.match(/^#?([0-9a-f]{6})$/i);
-        if (result) {
-            const text = result[1];
+        }else if (text && (text.length === 6 || text.length == 8) ) {
             const factor = 1 / 255;
             this.x = Number.parseInt(text.substr(0,2), 16) * factor;
             this.y = Number.parseInt(text.substr(2,2), 16) * factor;
             this.z = Number.parseInt(text.substr(4,2), 16) * factor;
-            this.w = alpha;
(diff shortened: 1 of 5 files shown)


gjcope:
This all looks good. I will test asap. I was not familiar with the custom validity support but seems like it could be useful in a number of places. I'm also wondering if it might be useful to have some sort of alpha enabled property in PropertyColor rather than relying on elementCount, but I'm not too concerned. Thanks for the PR!

gjcope:
Works great. Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-45 for team testing.

Not related to this PR specifically, but as I was attempting to copy a hex value from one input to another, I realized I had the initial impulse to click the text input to activate and then click and drag to select the text, which doesn't work. I don't think it's a big problem, but something to keep an eye on in case it becomes a bigger UX issue noted by more users.

sdumetz:
> Not related to this PR specifically, but as I was attempting to copy a hex value from one input to another, I realized I had the initial impulse to click the text input to activate and then click and drag to select the text, which doesn't work. I don't think it's a big problem, but something to keep an eye on in case it becomes a bigger UX issue noted by more users.

Would be great. It should be easy with drag&drop callbacks no? I'll try to do it when I have a chance.

gjcope:
I was just thinking drag to select active text like in a typical text input, but drag/drop could work too. You can actually drag/drop right now, but it zeroes out the source input. But again, not a big concern.

gjcope:
I spoke too soon, just noticed one small issue. The style is off for the color picker widgets in Explorer.
image

gjcope:
Some style fixes made here: https://github.com/Smithsonian/dpo-voyager/commit/c49c02fd727140a4cafb39947c247ea1d1cb71ab

PR #305, opened 2024-09-06
update-in-place for three r168

sdumetz:
I needed an update to three.js to test https://github.com/mrdoob/three.js/pull/27098 from r161 so I thought I might as well make a PR for it.

There was some uninteresting types rewrites, some frag/vert updates that I extracted from a diff of the physical shader from r158 to r168. I only used this in local tests so it might break in unexpected ways.


gjcope:
This is long overdue, thanks! Now I'm also interested in your use of WebGPURenderer ;)

PR #312, opened 2024-10-17
fix types for Publisher.emit

sdumetz:
Stumbled upon this working on something else and I thought I should patch it :

Fixes `Publisher.emit()` that returned `void` instead of `boolean` and accepted `any` values for both parameters.

Return value wasn't used outside of CHierarchy in a routine I'm not sure is even needed, but the signature is consistent with Node's [EventEmitter.emit](https://nodejs.org/api/events.html#emitteremiteventname-args) so we should keep it this way.


gjcope:
Thanks for the PR! Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-46

Issue #314, opened 2024-10-22
Camera position improperly linked to CVOrbitnavigation position


sdumetz:
In the settings tab if I edit the camera's transform, the changes are not propagated to CVOrbitNavigation (and in turn to the CameraController). Which means afterwards if I click to move the camera it will move relative to its old position, which is very much not what I want.

I see many ways this could be fixed but what seems most appropriate to me would be to make the camera's transform read-only to convey the existing fbinding from CVCameraController to CVCamera

We can achieve that using little new code with the existing `linkFrom` and `linkTo` mechanisms, and disable linked properties in the interface.

Is there a more elegant solution I'm not seeing?

gjcope:
Fixed by #315, merged in v0.46.0

PR #315, opened 2024-10-22
prevent interacting with linked Property

sdumetz:
Continuing my crusade against broken data bindings.

Properties that have their values linked from another prop should probably not be interactive.

Because the link is unidirectional, the value change is not propagated and it leads to inconsistencies.

The only case where this is currently visible is with the Lights transform, when `lightsFollowCam` is enabled. If I change this value, it's ignored the next time I move the camera.

This is a very minor annoyance but if this gets merged I intend to use such a binding to control the **transform** inputs of **CVCamera**, thus allowing the camera's transform setting to be locked-up, fixing #314.

See https://github.com/Smithsonian/dpo-voyager/commit/4bab99f17e9f7fee4ac8995444e7f19c4c088a43 for changes that would be required to link Cameras transform from CVOrbitNavigation.

### DIFF
--- a/source/client/ui/PropertyNumber.ts
+++ b/source/client/ui/PropertyNumber.ts
@@ -32,6 +32,9 @@ export default class PropertyNumber extends CustomElement
     @property({ type: String })
     name = "";
 
+    @property({ type: Boolean })
+    disabled :boolean = false;
+
     /**
      * Handles vector properties by specifying an array index.
      */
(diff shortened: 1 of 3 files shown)


gjcope:
Makes logical sense and seems to work well, thanks! Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-46

PR #316, opened 2024-10-23
save button collapse and smaller logo on small screens

sdumetz:
I tried to tackle those UX issues : 

- The "Download" button is proeminently shown, while seldom used afaik. 
- The "Save" button not saving the full scene "setup" is a source of confusion for many of our users.

I understand the need for a 2-stage save and while it is a source of complexity and confusion (I can think of no similar feature in other softwares), I don't think we'd be better off with a single save button and users blindly overwriting camera settings and the likes each time they edit something.

So if we can't get rid of it, I'd like to make this "capture setup" thing more discoverable by proposing a "save" dropdown that clearly show the save options that are available.

Using this it would be easy to further restrict options based on the "Mode", like what is done for "Standalone".

I also made the whole navbar a bit more compact for smaller screens but it's only marginally related.


gjcope:
There are a couple of things stylistically I would change with the reduced logo, but other than that everything seems functionally good.

Also not a big fan of the 2 stage save, but overwriting camera positions was a surprisingly common occurrence/complaint prior to that. Happy to go a different way if there is a suggestion that also avoids that issue. At a minimum with something like this, I think the vocabulary used should be the same in both 'capture' location so it's clear.

I want to get some more opinions on my end, but it seems like a positive direction. Thanks for the effort!

sdumetz:
> Also not a big fan of the 2 stage save, but overwriting camera positions was a surprisingly common occurrence/complaint prior to that.

That's my answer to everyone saying "there should only be one save button" so we are on the same line here.

I'd be very interested of the feedback you can gather on this. Maybe @kgschoueri may have some thought on this too? I know they are working a lot on the onboarding experience for non-technical users so anything we can do to improve how users save their changes would have a huge impact for them.

sdumetz:
Picking this up after way too much time, sorry.

I changed the logo styles to have the whole block float left on collapse. Maybe I should generalize it to the top bar logo so it collapses before the title starts clipping?

I applied your fix to wrap the caret + text together.

Changed the capture to "Save Scene State", I think it's a good label, we can always change it later if it's confusing. I made a custom icon trying to convey the button's meaning too.

Getting rid of taskbar scrolling would require either hiding the button labels or having the main taskbar collapse on small screens. The latter would probably work well-enough, but that's a separate change. Until then I think the scroll fallback is still better than having some buttons be unreachable.

Let me know what you think!

PR #318, opened 2024-10-23
deduplicate models bounding box computation

sdumetz:
Specifically [BufferGeometry.computeBoundingBox](https://threejs.org/docs/?q=Mesh#api/en/core/BufferGeometry.computeBoundingBox), which sets `BufferGeometry.boundingBox`, looks like it is not used anywhere.

The code uses [computeLocalBoundingBox](https://github.com/Smithsonian/dpo-voyager/blob/dff5bcf0425c2c4b0ec79c0ed3ef8afc2b30d793/libs/ff-three/source/helpers.ts#L106), using more or less the same algorithm but with additional tree traversal and matrix composition logic.

fwiw, each of those can take up to 100ms to compute on large-ish models so it can add up quickly. Ideally i'd totally skip this if the box can be retrived from the svx document or at least compute only the **Thumb** model's box and roll with it, but that'd be a more substantial change for another time.

### DIFF
--- a/source/client/io/ModelReader.ts
+++ b/source/client/io/ModelReader.ts
@@ -133,8 +133,6 @@ export default class ModelReader
                    material.map.colorSpace = SRGBColorSpace;
                 }
 
-                mesh.geometry.computeBoundingBox();
-
                 const uberMat = material.type === "MeshPhysicalMaterial" ? new UberPBRAdvMaterial() : new UberPBRMaterial();
 
                 if (material.flatShading) {


gjcope:
Is this really saving anything or just shifting the overhead?

When the scene bounds get recomputed in CVScene https://github.com/Smithsonian/dpo-voyager/blob/b352cbe6f19cda1bafd08e6f73e425a3b4b7ad89/source/client/components/CVScene.ts#L177
expandByObject is used. Looking at that source, if the bounding box is not set in the Three object/geometry, it's going to get called anyway for each object.

https://github.com/mrdoob/three.js/blob/dev/src/math/Box3.js#L192

gjcope:
Ah, yeah sorry I accidentally linked a proof-of-concept branch above where the precise bounds were needed. That could be another option, but obviously the optimization value would be different.

gjcope:
Merged with rc-48. Did some more testing on this and confirmed that it is just shifting the bounding box computation to the expandByObject function in the scene bounds calculation. But I also tested standard/standalone and single/multi object files and found no meaningful difference in the resulting scene bounds so I think it's fine to simplify and remove it. But let me know if you notice any issues.

PR #319, opened 2024-10-24
Eliminate forced reflows when rendering HTMLSprites

sdumetz:
`HTMLSprite.renderHTMLElement` and `HTMLSpriteGroup.render` were causing unnecessary forced reflows within their render loops.

skipping `getBoundingClientRect()` where possible solves it.

Basically free performance gain, about .5 ms per annotation.

### DIFF
--- a/libs/ff-three/source/HTMLSpriteGroup.ts
+++ b/libs/ff-three/source/HTMLSpriteGroup.ts
@@ -16,6 +16,14 @@ import HTMLSprite from "./HTMLSprite";
 
 export { HTMLSprite };
 
+interface IViewport{
+    x: number;
+    y: number;
+    width: number;
+    height: number;
+    overlay: HTMLElement;
+}
+
 /**
  * THREE 3D object, grouping a number of HTML sprites.
  */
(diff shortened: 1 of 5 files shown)


gjcope:
I really like this optimization, but it seems to have broken the annotation truncation functionality. I'll take a deeper look when I have a chance but see the screen shots below for an example of the result with this PR (top) and master (bottom) when panning an annotation to the top edge of the component element.

annotation_issue

libs/ff-three/source/HTMLSpriteGroup.ts
```diff
@@ -69,18 +77,19 @@ export default class HTMLSpriteGroup extends Object3D
...
      */
-    render(container: HTMLElement, camera: Camera)
+    render(viewport: IViewport, camera: Camera)
     {
         if (!this.visible) {
             return;
         }
-
+        //Only get bounds once to prevent forced reflows while looping
+        const bounds = new DOMRect(viewport.x, viewport.y, viewport.width, viewport.height);
```
gjcope, line 86: Still need to test a little more, but looks like this line should be `const bounds = viewport.overlay.getBoundingClientRect();`

sdumetz: Yes it should. I thought they were the same, and they are in voyager-explorer, but not in a story page. 

There seems to be no significant performance hit when introducing this call so everything should be all right.

gjcope:
Merged to rc-46. Thanks!

PR #320, opened 2024-10-28
Italian UI Translation 

kgschoueri:
I want to add the italian translations of the UI for the next voyager release. I'm also not sure if I managed to update the german as requested by one of our users.


gjcope:
Hi @kgschoueri , yes the German update was included in release v0.42.0 (https://github.com/Smithsonian/dpo-voyager/commit/b8d21678cc5c6b68063fa790ff0f7e64ee07b937)

It looks like this PR also has some documentation updates in it. Was that intentional? Maybe you added to a previous PR instead of creating a new one?

PR #321, opened 2024-10-30
make evaluateProps reentrant in all applications

sdumetz:
`ExplorerApplication.evaluateProps()` won't yield the same result when run twice, because `const props = this.props` only references the original object, so down the line when we get the queryString values, it changes the values stored in `this.props`, thus potentially changing the result of `evaluateProps()` in unexpected ways.

I think we should keep `this.props` unchanged or make it clear we intend to modify it by removing the `const props = this.props` assignment.

I prefer the first solution as it's a one-line change and allows calling evaluateProps after a change in the page's queryString.

### DIFF
--- a/source/client/applications/ExplorerApplication.ts
+++ b/source/client/applications/ExplorerApplication.ts
@@ -261,7 +261,7 @@ Version: ${ENV_VERSION}
 
     evaluateProps()
     {
-        const props = this.props;
+        const props = {...this.props};
         const manager = this.assetManager;
         const qs = new URL(window.location.href).searchParams;
         props.root = props.root || qs.get("root") || qs.get("r");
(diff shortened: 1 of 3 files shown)


gjcope:
Do you have an example that triggers an issue with the current implementation? I think your change sounds fine, just trying to understand the issue a little better. Thanks!

sdumetz:
It's an triggered when you try to control the scene from an external script doing something like this : 

```
let u = new URL(window.location.href);
u.searchParams.set("root","/scenes/foo");
voyager.reloadDocument();
```

Clearly not something voyager was made for, but it seems to work well enough with this small change, which is way better than adding a new public method just for this edge case.

gjcope:
Got it, thanks. Sounds good. Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-46

PR #323, opened 2024-11-09
Allow Docker image to work with non-root users, Docker streamlining

jbylsma:
npm commands run through the provided Docker image were failing to run when the files in the mounted /app folder weren't owned by root. This occurred from a confluence of two issues:

* npm was switching the executed user away from root
* nvm was inaccessible because /root is 700

npm spawns processes using the `@npmcli/promise-spawn` package. The current supported npm version, 8.9.14, uses a version of `@npmcli/promise-spawn` that will detect if npm is running as root. If root is detected, npm will use functionality from the `infer-owner` package to infer a non-root user through file ownership and ultimately execute the command with that user. This functionality is provided by the `infer-owner` package. This functionality was later removed from `@npmcli/promise-spawn` in <https://github.com/npm/promise-spawn/commit/a8b21fc66b31e3d724f6b91c77d8bc4f6172ed13>.

Separately, the provided Docker image is built and configured to run as root. As nvm defaults to installing in the current user's home directory, nvm was installed in `/root/.nvm`. Additionally, the Docker image's base, `ubuntu:20.04`, uses 700 permissions for `/root`.

Therefore, user-inferred commands spawned by npm would not have permissions to run commands in non-globally accessible `/root/.nvm`. By moving nvm to a globally accessible location, npm is able to spawn
user-inferred commands.

As part of this debugging, I streamlined the Dockerfile creation, which significantly sped up builds.


gjcope:
Thanks so much for the PR @jbylsma . Working perfectly on linux with a couple of tweaks (https://github.com/Smithsonian/dpo-voyager/tree/dev_docker_update) but running into a somewhat similar issue on Windows.

The build is failing as 'npm install' attempts to [write to package-lock.json](https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json) and by default write privileges are only set for the file owner (which seems to differ from whatever user npm is switching to). If I explicitly give write permissions to the 'user' group on the local file system, all is good, but this seems like an unnecessary step. Pinging here in case you have any ideas. Thanks!

gjcope:
Scratch the above, I was looking at it the wrong way. The same issue exists on other platforms if package-lock is not owned by the user. I think this is a totally reasonable expectation (unlike the previous needing to be owned by root) so I will just add a build note. Thanks again!

gjcope:
Merged with https://github.com/Smithsonian/dpo-voyager/tree/rc-46

Issue #254, opened 2024-02-27
Derivative quality "High" is always used


sdumetz:
Current derivative quality selection is based on a test over `gl.MAX_TEXTURE_SIZE` (as reported by [WebGLRenderer](https://threejs.org/docs/?q=renderer#api/en/renderers/WebGLRenderer)). This test uses static values to determine the best model quality.

I the values used (up to `4096` for **High** quality) made sense in 2019 when they were created but nowadays, 4096 is a more or less a baseline and everyone ends up with the **High** quality model.

I couldn't find any definitive source to back this up, but see [MDN recommendations for webGL](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices) or [those openGL capabilities statistics](https://feedback.wildfiregames.com/report/opengl/feature/GL_MAX_TEXTURE_SIZE#8192) for example. To get an idea, 4096 is supported by a Raspberry Pi 4.

Instead of updating those to a larger, more up to date value, which would break all existing scenes that may have been optimized to match (serving low quality content on reasonably capable devices), I'd try to design something more flexible.

Since the scene file is already designed to store texture sizes with assets (under `Derivative[x].assets[x].imageSize` ), it should be possible to select an appropriate quality based on actual size, instead of constants. This might not be ideal because it sets aside concerns of bandwidth and vertex count.

Another problem would be to compute the *actual* image size : Not only do we have to iterate over all models to check for individual sizes, we can't know beforehand if the model has just one diffuse or many maps (that might themselves have varying size, not matching the `imageSize` attribute).

Maybe some other implementation solved it in a better way?

I'm leaving this issue here to gather comments from anyone that has experience on the matter, implementations ideas or requirements not listed here.

gjcope:
There's a significant difference between what is supported and what is a good user experience for a model being rendered in-browser. 4k textures add up quickly when in-memory. We've seen significant load issues with as few as 3 fairly low-poly models with 4k textures (3 channels), especially with mobile browsers. I haven't seen anything to suggest that 4k is not still a good choice for 'high' quality.

That being said, how the quality selection happens can definitely be improved. We have talked about something bandwidth driven but haven't had the time to pursue it.

sdumetz:
[...]

PR #324, opened 2024-11-14
Dynamic LOD switching

sdumetz:
So there is this idea_ I worked on for the past few months that makes it possible to open a ~200 models scene with each model having a 4k diffuse (in `High` quality) on a mobile device. 

It's still a very rough prototype but I wanted to share it and begin discussions on what requirements we'd have to meet for this to get merged into voyager.

This branch temporarily includes https://github.com/Smithsonian/dpo-voyager/pull/266 to allow more camera mobility  and a lot of performance patches that should be (or have been) extracted to their own PRs. It also needs  to be rebased on Voyager's latest release. I'll do all this down the line as I clean everything up.

A working example is visible there: 

https://lodtest.ecorpus.holusion.net/ui/scenes/Lascaux_retake2/view

Same model, without LOD (everything is squashed together and downscaled to a 8k diffuse map) is publicly available [on the french ministry of culture website](https://archeologie.culture.gouv.fr/fr/media_externe?src=aHR0cHM6Ly90aGliYXVsdGd1aWxsYXVtb250LmdpdGh1Yi5pby9sYXNjYXV4M2QvMjAyNC0wNS0xNC1sYXNjYXV4M0RfRlIuaHRtbA==&token=fc131ee67240b9f8108015a4b4ec7c843a52fd36) and takes longer to load initially, has a lot of blur and doesn't even work on most mobile devices.

I also have a simpler demo example that should clearly show what is happening: 

https://lodtest.ecorpus.holusion.net/ui/scenes/simple_room/view

## Reasoning

I started with 2 facts :
 - as stated in https://github.com/Smithsonian/dpo-voyager/issues/254 even lower-end devices are fully capable of showing High quality  (4k textures) derivatives and there is no real point in trying to serve larger models even for high-end devices
 - Quality is selected at the scene level. Voyager has the ability to use heterogeneous quality internally but no way to exploit this.

My objective was to be able to handle spatially large models. Things like a large painting, multi-figure sculpture or a whole room. In such an environment, I want what's right under my nose to have a very high texture resolution (abt. 4k for something filling the whole screen is a good upper bound) but on the other hand I don't care about the resolution of pretty much everything else.

Having a scene with 10s  of objects, each with 4k diffuses tends to overload even desktop computers with low end discrete GPUs. I want to be able to have 100s of them. 

## How it works

I repurposed the existing `Derivative` system to dynamically switch models quality depending on the "perceived importance".

Pretty much everything happens in the new [CVDerivativeController](https://github.com/Smithsonian/dpo-voyager/blob/bbc8efb6d7c1a0fa095e3a10c92e8e8dda0df284/source/client/components/CVDerivativesController.ts) class. The scene is initially loaded with `Thumb`quality (as it was before). Then the controller regularly sorts every model in the scene by importance (on-screen size and angle to camera are taken into account) and increments the quality of the most important model(s). If/When too much models are upgraded, it downgrades models as needed starting with the least important ones.

## Heuristics

_Tradeofs that had to be made. Can and should be adjusted._

### Performance

Estimating a performance budget is hard (see  https://github.com/Smithsonian/dpo-voyager/issues/254). Even if we could reliably determine a device's graphic power (we can't), we can't know if the user wants us to use 100% of his computer's power.

I settled somewhat arbitrarily on an initial budget of 2 `High` resolution models that is then further decreased if we detect a low-end device (CPU count, RAM size, is a mobile device).

### Model weight

Models are weighted based on their **size** in camera space, their **distance** to the camera and their **angle** relative to the  `Camera.forward`.

The Distance modifier is somewhat redundant with size but we kinda want far away models to be less important even if they are very big. In fact most LOD algorithms are purely distance-based.

### DIFF
--- a/libs/ff-graph/source/components/CPulse.ts
+++ b/libs/ff-graph/source/components/CPulse.ts
@@ -19,6 +19,8 @@ export interface IPulseContext extends IUpdateContext
     secondsElapsed: number;
     secondsDelta: number;
     frameNumber: number;
+    tickUpdated: boolean;
+    tockUpdated: boolean;
 }
 
 /**
@@ -53,7 +55,6 @@ export default class CPulse extends Component
     private _secondsStopped: number;
     private _animHandler: number;
     private _pulseEvent: IPulseEvent;
-    private _tockUpdated = false;
 
     constructor(node: Node, id: string)
     {
@@ -66,7 +67,9 @@ export default class CPulse extends Component
             time: new Date(),
             secondsElapsed: 0,
             secondsDelta: 0,
-            frameNumber: 0
+            frameNumber: 0,
+            tickUpdated: false,
+            tockUpdated: false,
         };
 
         this._secondsStarted = Date.now() * 0.001;
@@ -118,14 +121,14 @@ export default class CPulse extends Component
         outs.frame.setValue(context.frameNumber);
 
         // execute tick
-        const tickUpdated = this.system.graph.tick(context);
+        context.tickUpdated = this.system.graph.tick(context);
 
         // emit pulse event
-        _pulseEvent.systemUpdated = tickUpdated || this._tockUpdated;
+        _pulseEvent.systemUpdated = context.tickUpdated || context.tockUpdated;
         this.emit<IPulseEvent>(_pulseEvent);
 
         // execute tock
-        this._tockUpdated = this.system.graph.tock(context);
+        context.tockUpdated = this.system.graph.tock(context);
     }
 
     protected onAnimationFrame()
--- a/libs/ff-scene/source/components/CRenderer.ts
+++ b/libs/ff-scene/source/components/CRenderer.ts
@@ -6,7 +6,7 @@
  */
 
 import { Mesh } from "three";
-import * as constants from "three/src/constants";
+import * as constants from "three/src/constants.js";
 
 import Component, { Node, ITypedEvent, types } from "@ff/graph/Component";
 import CPulse, { IPulseEvent } from "@ff/graph/components/CPulse";
@@ -66,6 +66,7 @@ export default class CRenderer extends Component
     static readonly outs = {
         maxTextureSize: types.Integer("Caps.MaxTextureSize"),
         maxCubemapSize: types.Integer("Caps.MaxCubemapSize"),
+        framerate: types.Integer("Renderer.Framerate"),
     };
 
     ins = this.addInputs(CRenderer.ins);
@@ -221,7 +222,9 @@ export default class CRenderer extends Component
 
             this.views.forEach(view => {
                 if(!view.renderer.xr.isPresenting) {
+                    const start = (performance || Date).now();
                     view.render();
+                    this.outs.framerate.value = this.outs.framerate.value *0.9 + 0.1*1000/((performance || Date).now() -start);
                 }
             });
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -214,9 +214,9 @@
             }
         },
         "node_modules/@types/estree": {
-            "version": "1.0.1",
-            "resolved": "https://registry.npmjs.org/@types/estree/-/estree-1.0.1.tgz",
-            "integrity": "sha512-LG4opVs2ANWZ1TJoKc937iMmNstM/d0ae1vNbnBvBhqCSezgVUOzcLCqbI5elV8Vy6WKwKjaqR+zO9VKirBBCA=="
+            "version": "1.0.6",
+            "resolved": "https://registry.npmjs.org/@types/estree/-/estree-1.0.6.tgz",
+            "integrity": "sha512-AYnb1nQyY49te+VRAVgmzfcgjYS91mY5P0TKUDCLEM+gNnA+3T6rWITXRLYCpahpqSQbN5cE+gHpnPyXjHWxcw=="
         },
         "node_modules/@types/html-minifier-terser": {
             "version": "6.0.0",
@@ -236,9 +236,9 @@
             "dev": true
         },
         "node_modules/@types/node": {
-            "version": "16.18.108",
-            "resolved": "https://registry.npmjs.org/@types/node/-/node-16.18.108.tgz",
-            "integrity": "sha512-fj42LD82fSv6yN9C6Q4dzS+hujHj+pTv0IpRR3kI20fnYeS0ytBpjFO9OjmDowSPPt4lNKN46JLaKbCyP+BW2A=="
+            "version": "16.18.122",
+            "resolved": "https://registry.npmjs.org/@types/node/-/node-16.18.122.tgz",
+            "integrity": "sha512-rF6rUBS80n4oK16EW8nE75U+9fw0SSUgoPtWSvHhPXdT7itbvmS7UjB/jyM8i3AkvI6yeSM5qCwo+xN0npGDHg=="
         },
         "node_modules/@types/stats.js": {
             "version": "0.17.2",
@@ -2378,18 +2378,6 @@
                 "handlebars": ">= 1.3.0 < 5"
             }
         },
-        "node_modules/has": {
-            "version": "1.0.3",
-            "resolved": "https://registry.npmjs.org/has/-/has-1.0.3.tgz",
-            "integrity": "sha512-f2dvO0VU6Oej7RkWJGrehjbzMAjFp5/VKPp5tTpWIV4JHHZK1/BxbFRtf/siA2SWTe09caDmVtYYzWEIbBS4zw==",
-            "dev": true,
-            "dependencies": {
-                "function-bind": "^1.1.1"
-            },
-            "engines": {
-                "node": ">= 0.4.0"
-            }
-        },
         "node_modules/has-flag": {
             "version": "3.0.0",
             "resolved": "https://registry.npmjs.org/has-flag/-/has-flag-3.0.0.tgz",
@@ -2694,12 +2682,15 @@
             "integrity": "sha512-NcdALwpXkTm5Zvvbk7owOUSvVvBKDgKP5/ewfXEznmQFfs4ZRmanOeKBTjRVjka3QFoN6XJ+9F3USqfHqTaU5w=="
         },
         "node_modules/is-core-module": {
-            "version": "2.8.0",
-            "resolved": "https://registry.npmjs.org/is-core-module/-/is-core-module-2.8.0.tgz",
-            "integrity": "sha512-vd15qHsaqrRL7dtH6QNuy0ndJmRDrS9HAM1CAiSifNUFv4x1a0CCVsj18hJ1mShxIG6T2i1sO78MkP56r0nYRw==",
+            "version": "2.16.0",
+            "resolved": "https://registry.npmjs.org/is-core-module/-/is-core-module-2.16.0.tgz",
+            "integrity": "sha512-urTSINYfAYgcbLb0yDQ6egFm6h3Mo1DcF9EkyXSRjjzdHbsulg01qhwWuXdOoUBuTkbQ80KDboXa0vFJ+BDH+g==",
             "dev": true,
             "dependencies": {
-                "has": "^1.0.3"
+                "hasown": "^2.0.2"
+            },
+            "engines": {
+                "node": ">= 0.4"
             },
             "funding": {
                 "url": "https://github.com/sponsors/ljharb"
@@ -3284,12 +3275,12 @@
             "integrity": "sha512-Jsjnk4bw3YJqYzbdyBiNsPWHPfO++UGG749Cxs6peCu5Xg4nrena6OVxOYxrQTqww0Jmwt+Ref8rggumkTLz9Q=="
         },
         "node_modules/mkdirp": {
-            "version": "0.5.5",
-            "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.5.tgz",
-            "integrity": "sha512-NKmAlESf6jMGym1++R0Ra7wvhV+wFW63FaSOFPwRahvea0gMUcGUhVeAg/0BC0wiv9ih5NYPB1Wn1UEI1/L+xQ==",
+            "version": "0.5.6",
+            "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.6.tgz",
+            "integrity": "sha512-FP+p8RB8OWpF3YZBCrP5gtADmtXApB5AMLn+vdyA+PyxCjrCs00mjyUozssO33cwDeT3wNGdLxJ5M//YqtHAJw==",
             "dev": true,
             "dependencies": {
-                "minimist": "^1.2.5"
+                "minimist": "^1.2.6"
             },
             "bin": {
                 "mkdirp": "bin/cmd.js"
@@ -4518,9 +4509,9 @@
             "integrity": "sha512-77DZwxQmxKnu3aR542U+X8FypNzbfJ+C5XQDk3uWjWxn6151aIMGthWYRXTqT1E5oJvg+ljaa2OJi+VfvCOQ8w=="
         },
         "node_modules/punycode": {
-            "version": "2.1.1",
-            "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.1.1.tgz",
-            "integrity": "sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==",
+            "version": "2.3.1",
+            "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.3.1.tgz",
+            "integrity": "sha512-vYt7UD1U9Wg6138shLtLOvdAu+8DsC/ilFtEVHcH+wydcSpNE20AfSOduf6MkRFahL5FY7X1oU7nKVZFtfq8Fg==",
             "engines": {
                 "node": ">=6"
             }
@@ -4765,13 +4756,17 @@
             "integrity": "sha512-KigOCHcocU3XODJxsu8i/j8T9tzT4adHiecwORRQ0ZZFcp7ahwXuRU1m+yuO90C5ZUyGeGfocHDI14M3L3yDAQ=="
         },
         "node_modules/resolve": {
-            "version": "1.20.0",
-            "resolved": "https://registry.npmjs.org/resolve/-/resolve-1.20.0.tgz",
-            "integrity": "sha512-wENBPt4ySzg4ybFQW2TT1zMQucPK95HSh/nq2CFTZVOGut2+pQvSsgtda4d26YrYcr067wjbmzOG8byDPBX63A==",
+            "version": "1.22.9",
+            "resolved": "https://registry.npmjs.org/resolve/-/resolve-1.22.9.tgz",
+            "integrity": "sha512-QxrmX1DzraFIi9PxdG5VkRfRwIgjwyud+z/iBwfRRrVmHc+P9Q7u2lSSpQ6bjr2gy5lrqIiU9vb6iAeGf2400A==",
             "dev": true,
             "dependencies": {
-                "is-core-module": "^2.2.0",
-                "path-parse": "^1.0.6"
+                "is-core-module": "^2.16.0",
+                "path-parse": "^1.0.7",
+                "supports-preserve-symlinks-flag": "^1.0.0"
+            },
+            "bin": {
+                "resolve": "bin/resolve"
             },
             "funding": {
                 "url": "https://github.com/sponsors/ljharb"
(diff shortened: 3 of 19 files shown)


gjcope:
This is exciting, thanks so much! It may take a minute to get to, but I'm looking forward to testing it out.

gjcope:
@sdumetz Nice work. In terms of what we need to get it merged, I think a step that would be helpful (for me) would be to have a PR without #266 to make things a little clearer. Then I think we will need at least a few different test cases outside of Lascaux (though that one works quite well) which we can help with on our end. I think your general approach is great and clearly well thought out. There are a few corner cases to be considered, but I see you've got many of them already commented.

sdumetz:
Cleaned it up, removed #266 (I'll still include it in my test builds because otherwise the cave scenes are totally unusable)

We pushed it to a ~700 tiles model where it unsurprisingly showed its limits. This is unfortunately about 1/10th of what we need to have a full tour of the [Chauvet Cave](https://archeologie.culture.gouv.fr/chauvet/en).

I plan to add a way to totally unload some models from memory once they are farther out than the camera's `far` distance. This would also speed up initial loading by not bothering with out-of-view models initially.

Aside from caves we have access to high-quality models of XIII<sup>th</sup> century stone sculptures fragments so I plan on creating scenes with them shortly,  but any samples you have would be welcomed, this definitely need to be tested in more diverse cases.

gjcope:
Dug into this some with some simple test cases and found a couple of issues with the visibleSize calculation.

- Looks like the object pose is not being taken into account when applying transforms to the local bounding box
- The bounds check comparing the z-component of the projected coordinates with the near plane won't work because the projected depth is normalized and the near plane is not. So anything with a near plane < 1 (we have a number of those things) will end up with a null resulting bounds.

Suggested patches here: https://github.com/Smithsonian/dpo-voyager/commit/d0a31 but still needs testing so feel free to address otherwise.

Generally, it would be nice to not have it evaluate every frame and be driven by camera movement (or load), but I think that can be an optimization. Planning to do more testing soon.

sdumetz:
The angle computation was aimed at prioritizing things that are centered on-screen. It makes a lot of sense in interior spaces when the "tiles" are generally smalled than the camera"s frustrum. A lot less when zoomed-in on models like the suit. I suppose the `distance` computation will have the same limitation.

Ideally I'd like to compute those against the "best-case-point" from `_localBox` but not sure how to compute this.

> I'm seeing some 'Violation' warnings like this: https://discourse.threejs.org/t/violation-requestanimationframe-handeler-took-xx-ms/49567

Curious what contributes meaningfully to this? Is the scene too long to render or does the tree traversal meaningfully contribute to this? Wrapping the `tock` in a `setTimeout` to move it outside of `requestAnimationFrame`'s microtask is an option but I think it is more likely to be hiding the issue than fixing it.

If it's not the render or `CVDerivativesController.tock()` that's causing it, we should probably move the culprit outside of `requestAnimationFrame`'s scope.

> This scene also has a couple of objects that only have a 'high' derivative. This can create a condition where they are flagged for quality change every frame but are unable to do so. Not a serious issue, but it does result in "derivative quality 'XXX' not available, using higher quality" message being spammed to the console which can eventually affect browser performance.

Maybe once everything is settled we move this message to _somewhere_ where it's triggered only when dynamic LOD is disabled and we are actually trying to load a specific quality and are not able to do so.

Pushing an algorithm change alongside your patch and a rebase on `master`,  to more aggressively unload higher-res models when they are no longer needed. Hopefully this is also more readable though I'm still not very proud of my work in this matter.

gjcope:
> The angle computation was aimed at prioritizing things that are centered on-screen. It makes a lot of sense in interior spaces when the "tiles" are generally smalled than the camera"s frustrum. A lot less when zoomed-in on models like the suit. I suppose the `distance` computation will have the same limitation.
> 

Got it. I think the updated calculation makes sense for this.

> Curious what contributes meaningfully to this? Is the scene too long to render or does the tree traversal meaningfully contribute to this? Wrapping the `tock` in a `setTimeout` to move it outside of `requestAnimationFrame`'s microtask is an option but I think it is more likely to be hiding the issue than fixing it.

Good question. I'll see if I can narrow it down.

With the current update I'm seeing a regression testing the space suit example above. Just zooming in from the default view results in the assets (same quality) being repeatedly reloaded over and over. Looking into it, but tough to debug with the console spamming.

gjcope:
The issue with the astronaut scene is that when you zoom in close, every asset returns 'High' for getQuality. So the first two downgrade clauses get triggered but are ineffective since the qualityRequest is still 'High' and they fall into the 'contingency' clause to actually get downgraded. Since nothing has changed from a camera perspective, getQuality will still return 'High' on the following pass and so it oscillates back and forth. 

I haven't sorted out yet why this didn't happen with the previous implementation since the relativeSize calculation doesn't seem to have changed.

sdumetz:
Should hopefully be fixed in [83bb987](https://github.com/Smithsonian/dpo-voyager/pull/324/commits/83bb987e52a3daabe228a21c4a6517103d425391).

gjcope:
Looks like it still cycles back and forth between upgrade/downgrade for the astronaut example.

gjcope:
This update definitely seems to have fixed the instability, thanks! It was easily reproducible with the last version.

I'll experiment with some more tweaking. Right now what I'm seeing is once you zoom in and establish a 'medium' quality of all pieces, navigation only results in a quality change for the same piece (p1), up and down, unless you zoom out.

sdumetz:
The suit can behave weirdly because its pieces are not very "sphere-like" or "tile-like", which is are the way you'd ideally cut your pieces when designing a scene for LOD. Not that it makes it a bad test scene : it is good to test some edge cases (and it has :wink: ) and we should expect this kind of scenes to work well-enough. But we can't expect perfectly accurate weight estimates for such volumes. After all, our main source of weight is the bounding box, which is a bad approximation for meshes that have multiple disjoint bodies.

Another thing to keep in mind is we are not trying to be super aggressive with unloads because once something is loaded into memory we kinda don't want to have to load a replacement (even a lower Q one) unless we need to free up the memory: Network + GPU transfer times are by far our worse problem in most cases.

Unless of course too much high quality models make the scene render too slow or crash.

My method to determine this "budget" is one of the things that I know I'll need to revisit once in the future with a bit more "real world" usage knowledge.

gjcope:
My current testing is actually showing the weighting to be as expected with the space suit scene. The issue seems to be more with how the downgrades are handled. The downgrades I would expect are requested, but not processed. It's fairly easy to get it in a state where it is overbudget and stays that way with no downgrades. Still digging into it.

sdumetz:
I pushed an update that should fix this. I'm cancelling as many upgrades as necessary before contingency downgrades.

I also added a FPS moving average calculation to further reduce budget if FPS is too low. I am trying to find more cases where render performance is actually a problem. with the current state of things to see if I can improve the budget calculation. I managed to grab an old 5th gen i5 laptop that should make a good performance baseline.

gjcope:
Nice work, this looks like a big improvement to the scheme! I think there are a few remaining issues/concerns but seems close to pushing to a wider audience for testing. Issues/questions below:

1. The loading & interaction indicators are broken again. See https://github.com/Smithsonian/dpo-voyager/tree/dev-lod-tweak for a partial fix, but fundamentally we need a better way of defining the 'initial load' period for it to work. Previously we had an explicit load of the 'lowest' derivative and then the 'target' derivative, so there was good point to catch this. Now it is a little more nebulous, so I'm not sure exactly how we track this. This also affects capturing load time stats which are important for our metrics.
2. The on-the-fly streaming can have a decent impact on navigation (stuttering, freezing). Seems like we are still stuck on some synchronous activity somewhere? Haven't been able to dig deeper into that one yet.

Really great job, I'm excited to have this feature available!

sdumetz:
> The loading & interaction indicators are broken again.
Now that I think about it we are still loading _most_ objects as thumbnails before doing any upgrade job so it might make sense to delay any upgrade until that "initial load" event fires. This would have the added benefit of preventing this code from breaking the initial load any other day, which I very much see it do in its current state. I'll check if that would work first and if not I'll try to find a robust alternative.

> The on-the-fly streaming can have a decent impact on navigation (stuttering, freezing). Seems like we are still stuck on some synchronous activity somewhere? Haven't been able to dig deeper into that one yet.

A lot of things are still synchronous. Just as interleaved as possible with the normal rendering flow to reduce delayed frames. The upload of decoded textures to the GPU in particular can take a long time and it is synchronous by design in WebGL. I'm not even sure if WebGPU addresses that, and even then I don't expect support and backward compatibility to be good enough to use it for at least a few years.

KTX textures tends to upload faster due to their lower memory footprint (they are uploaded in a compressed state).

Aside from that, most things are asynchronous now : Image decoding should happen off the main thread, shader compilation should be parallelized (at least when [KHR_parallel_shader_compile](https://developer.mozilla.org/en-US/docs/Web/API/KHR_parallel_shader_compile) is supported afaik, so not in firefox). 

I'm not sure about the geometry upload. I _think_ it's inherently synchronous like texture transfer, but not 100% sure about it. In any case there is nothing we can do. meshopt compression is _generally_ faster to decompress than draco (both seems to be blocking), so it's a tradeoff. In cases where we really don't care about bandwidth, we are disabling compression to get things a bit faster.

We might need some sort of controls to tune (disable?) this in cases where it actually makes a scene worse. There is no way to guarantee for any collection of models to be able to load seamlessly on all devices.

gjcope:
> A lot of things are still synchronous. Just as interleaved as possible with the normal rendering flow to reduce delayed frames. The upload of decoded textures to the GPU in particular can take a long time and it is synchronous by design in WebGL. I'm not even sure if WebGPU addresses that, and even then I don't expect support and backward compatibility to be good enough to use it for at least a few years.

Fair point. It's anecdotal, but I think what jumped out to me is how much more disruptive it seems than, for instance, Sketchfab's streaming LOD so I don't know that it's entirely up to inherent WebGL limitations. But this is just a first pass and I think the performance is good for that.

sdumetz:
In most cases, seamless LOD is way more dependant upon the data encoding than the LOD selection algorithm. In fact most LOD systems are trivially simple (just stick a bunch of quality thresholds by distance to camera and be done with it). The most complete LOD selection algorithm that I know of is made by [3D tiles](https://github.com/CesiumGS/3d-tiles), and even that does nothing that would really help us.

As I understand it, Unreal's [Nanite](https://dev.epicgames.com/documentation/en-us/unreal-engine/nanite-virtualized-geometry-in-unreal-engine)+[Virtual Textures](https://dev.epicgames.com/documentation/en-us/unreal-engine/virtual-texturing-in-unreal-engine) and Sketchfab's [MASSIVE](https://sketchfab.com/blogs/community/introducing-massive-by-sketchfab-alpha/) are using a "streamable" file format: models sliced into chunks so tiny they make the synchronous upload steps individually negligible.

One of the additional simplifications Sketchfab made to achieve this is (was?) to ditch textures completely and only show vertex colors. Additionally they seem to use one of the newer, proprietary "streaming-enabled" file formats. Somewhat akin what [POTREE](https://github.com/potree/potree) does, but for vertex clusters.

Unreal's streaming technology is 50% dark magic at this point, I understand the basics of streaming a mesh chunked into vertex clusters (as long as you keep the cluster's boundaries, you should have no visible seams) but doing that while not messing-up your UV map is quite a feat.

AFAIK glTF has no [current or planned extension](https://github.com/KhronosGroup/glTF/blob/main/extensions/README.md) for mesh or texture streaming. I don't think that should make glTF a bad candidate for LOD streaming: It just shifts the burden of making chunks that could reasonably be loaded within a frame onto the author.

Open Source web-enabled streaming formats are developped as we speak but I don't know how ready they are for general availability. None of them are currently included in THREE's addons yet.

gjcope:
If you have a chance to review and feel like this is out of draft status it would be great to get it into this month's release.

sdumetz:
Hi,

I think we're nearly there.

I'm not entirely pleased with how the relative size computation work so I had @Inao0 take a look at it to see if we can improve it.
One way or another, this should be done in a few days.

Then I think there is two issues we never really discussed : 
- Do we want this feature to be optional or always on? if optional, what should be the default?
- Are you OK with how it currently interact with the **derivatives** task? I generally don't use it so I didn't bother with it but I think it will override whatever quality you select.

gjcope:
> * Do we want this feature to be optional or always on? if optional, what should be the default?

We were thinking off by default for the initial release. On our end we will enable it for a few scenes that could really benefit from it and gather some more real-world feedback. Then ideally we make it default  on in a future release.

> * Are you OK with how it currently interact with the **derivatives** task? I generally don't use it so I didn't bother with it but I think it will override whatever quality you select.

That's a really great question, I had completely overlooked that. Since our derivative creation pipeline is fully automated, the derivatives task is key for our QC step to make sure no processing issues occurred.

sdumetz:
Thinking about it, we just have to toggle its "enabled" property when we activate the `CVDerivativesTask`. I implemented it in 1384b3c.

About that, any reason why `this.startObserving();` and `this.stopObserving()` are not called in the base `CVTask`class declaration? It made CVDerivativesTask ignore most of its configuration options, which seems wrong?

Anyways, I think that should work. I'll clean things up, toggle the property to default off and then we should be ready for a final review.

gjcope:
> Thinking about it, we just have to toggle its "enabled" property when we activate the `CVDerivativesTask`. I implemented it in [1384b3c](https://github.com/Smithsonian/dpo-voyager/commit/1384b3c5809d59e1f629948aed30197ba79a5485).
> 

Great, thanks!

> About that, any reason why `this.startObserving();` and `this.stopObserving()` are not called in the base `CVTask`class declaration? It made CVDerivativesTask ignore most of its configuration options, which seems wrong?
> 

The configuration option set in the constructor? I'm not sure why that would be related to the observer, I'll have to take a look. I can't speak to the original motivation, but my guess would be because not all tasks are dependent on node/document changes. But I'm not opposed to handling it that way if it seems cleaner, but would also need to cleanup all the other tasks to remove redundancy on that front.

> Anyways, I think that should work. I'll clean things up, toggle the property to default off and then we should be ready for a final review.

Awesome. Excited to see this in the wild. Great work as always.

sdumetz:
Updated with a brand new relative size computation  algorithm that fares better in some scenes where we were seeing suboptimal results.

I added an optimization to run the computations only when the scene changes. Cleaned up a bit too.

I'm not sure about the names here : The feature is titled `Derivatives selection > LOD > enabled` in the **Settings** task and `setups[0].derivatives.enabled` in the scene file. None of those strikes me as particularly good. I'd gladly take suggestions here :smile:

gjcope:
> Updated with a brand new relative size computation algorithm that fares better in some scenes where we were seeing suboptimal results.

I'm seeing mixed results here. It seems more performant as the swaps aren't happening as often, but in one of our key use cases, the spacesuit linked in this thread, some models never make it to high-res. Particularly, the central piece of the suit with all the patches and lettering only gets to medium regardless of its size on screen compared to the other pieces. The previous algorithm swapped more like I would expect, but did swap rapidly as you moved around. I understand there's a balance with that, but if I piece doesn't load the high-res version when the other objects are out of frame it seems problematic.

This is a corner case, but I also noticed that two non-visible pieces in that scene are getting set as high from the outset. This could be an opportunity for optimization at some point.
> 
> I added an optimization to run the computations only when the scene changes. Cleaned up a bit too.

Nice work!

> 
> I'm not sure about the names here : The feature is titled `Derivatives selection > LOD > enabled` in the **Settings** task and `setups[0].derivatives.enabled` in the scene file. None of those strikes me as particularly good. I'd gladly take suggestions here 😄

Yeah, I see your point. Maybe 'Optimization' instead of 'Derivative selection'? That way future optimizations of different types could be enabled/disabled in that section. For the setup, not sure, but it does seem a little misleading to call it 'derivatives' since the actual derivatives exist elsewhere.

Inao0:
Hello @gjcope,

I tried to reproduce your issue. I used the model of the suit that you provided earlier in the discussion (https://3d-api.si.edu/content/document/d8c63ba6-4ebc-11ea-b77f-2e728ce88125/document.json).

From what I could see, this is what is happening :
- The budget is limited on the device you tried it on, meaning not all the high-quality textures can be loaded at the same time. It appears to be limited at 3 high textures max if you do not get central part in high quality.
- When you zoom in from the starting position, parts of the suit with a larger bounding box take the priority, leading to the side parts and the interior being loaded first. (For example, zooming in from the top of the suit loads the central part sooner.)
- When you are close enough for the central part to load in high quality, no more budget is available to load additional textures.

> Particularly, the central piece of the suit with all the patches and lettering only gets to medium regardless of its size on screen compared to the other pieces. The previous algorithm swapped more like I would expect, but did swap rapidly as you moved around. I understand there's a balance with that, but if a piece doesn't load the high-res version when the other objects are out of frame it seems problematic.

I hear this is frustrating, but the other objects are actually not out of frame when reasoning with the bounding boxes of the current pieces of the suit. The "Apollo" patch in particular is within the bounding boxes of the right side, central part, left side (due to the patch on the back of the right side included in it) and interior.
Due to the high amount of overlap of bounding boxes in the chest area, the current algorithm cannot downgrade anything when zooming in on the patch. It seems that the shapes of the different parts of this model are not very well suited to computing dynamic levels of details, and, sadly, especially not in this area of the model. Any chance it could be split more / differently ?

The new relative size computation algorithm works better especially when you have objects across the camera plan, reducing flickering in requested quality. Hence the difference in quality stability you observed and (maybe) why it was loading the central part in high quality before on the same computer.

gjcope:
> Hello @gjcope,
> 
> I tried to reproduce your issue. I used the model of the suit that you provided earlier in the discussion (https://3d-api.si.edu/content/document/d8c63ba6-4ebc-11ea-b77f-2e728ce88125/document.json).
> 
> From what I could see, this is what is happening :
> 
> * The budget is limited on the device you tried it on, meaning not all the high-quality textures can be loaded at the same time. It appears to be limited at 3 high textures max if you do not get central part in high quality.
> * When you zoom in from the starting position, parts of the suit with a larger bounding box take the priority, leading to the side parts and the interior being loaded first. (For example, zooming in from the top of the suit loads the central part sooner.)
> * When you are close enough for the central part to load in high quality, no more budget is available to load additional textures.

I get this, but I still don't think it is the desired behavior. I haven't had a chance to dig into the new algorithm as much as I did with the last, but from my perspective this is a regression in this aspect. Even if the bounding boxes are overlapping, of the objects in frame the one I'm closest to should get priority, no? And if there is no more of the budget available, something should be downgraded?

> 
> > Particularly, the central piece of the suit with all the patches and lettering only gets to medium regardless of its size on screen compared to the other pieces. The previous algorithm swapped more like I would expect, but did swap rapidly as you moved around. I understand there's a balance with that, but if a piece doesn't load the high-res version when the other objects are out of frame it seems problematic.
> 
> I hear this is frustrating, but the other objects are actually not out of frame when reasoning with the bounding boxes of the current pieces of the suit. The "Apollo" patch in particular is within the bounding boxes of the right side, central part, left side (due to the patch on the back of the right side included in it) and interior. Due to the high amount of overlap of bounding boxes in the chest area, the current algorithm cannot downgrade anything when zooming in on the patch. It seems that the shapes of the different parts of this model are not very well suited to computing dynamic levels of details, and, sadly, especially not in this area of the model. Any chance it could be split more / differently ?
> 
> The new relative size computation algorithm works better especially when you have objects across the camera plan, reducing flickering in requested quality. Hence the difference in quality stability you observed and (maybe) why it was loading the central part in high quality before on the same computer.

The majority of our scenes (> 3000)  are object-centric experiences where you are orbiting the objects and zooming in/out. I made a test scene that seems like it should work well with your parameters, but it still does not. Four objects of varying sizes in a row with no overlapping bounding boxes. The first 3 objects (l-to-r visually) load in at high resolution and the last (helmet) stays at thumb (in reality low, because there is no thumb available). No matter how you navigate the scene, zooming in on the helmet such that there are no other bounding boxes in frame, the quality of the helmet never changes. Given this and the spacesuit, I think there is an issue with how downgrades are being managed. I will dig into it more when I have the cycles.

image

Inao0:
I agree that this is definitely not the expected behavior when objects are placed side by side in a scene. We did not encounter this issue when testing with our own scenes, thank you for pointing it out.
The part I worked on previously was only changing the weight calculation.
I took a closer look at the upgrade/downgrade part of the system. There was indeed no mechanism that allowed parts to be downgraded to make room for parts with higher weights/priorities. The last commit implements a solution. 
The switch happens if the weight of the part to be upgraded is 1,5 greater than the weights of the parts that need to be downgraded to make room. This can be adjusted if necessary. This threshold prevents switching back and forth too much.
I think this fixes your issue with the patches on the suit, although you have to get very close.

> Has this been investigated? Commented out, it causes null bounds to be written to the scene on save.
I haven't had time to look at this yet. I'll check it next week.

gjcope:
> I agree that this is definitely not the expected behavior when objects are placed side by side in a scene. We did not encounter this issue when testing with our own scenes, thank you for pointing it out. The part I worked on previously was only changing the weight calculation. I took a closer look at the upgrade/downgrade part of the system. There was indeed no mechanism that allowed parts to be downgraded to make room for parts with higher weights/priorities. The last commit implements a solution. The switch happens if the weight of the part to be upgraded is 1,5 greater than the weights of the parts that need to be downgraded to make room. This can be adjusted if necessary. This threshold prevents switching back and forth too much. I think this fixes your issue with the patches on the suit, although you have to get very close.

Nice work. This definitely fixes the space suit issue. You do have to get close for the patch section, but I think that's an acceptable artifact of how it's divided up, as you previously mentioned. When I navigate around it I see the behavior I would expect.

For some reason I'm still not seeing the other example (objects in a line) working. It still upgrades 3 objects to high and never the 4th, regardless of navigation. I guess maybe this is related to the threshold you mentioned? I will try experimenting with it.

Overall looks like a very positive improvement. Thanks!

> 
> > Has this been investigated? Commented out, it causes null bounds to be written to the scene on save.
> 
> I haven't had time to look at this yet. I'll check it next week.

Sounds good. I think that and the hardcoded 'ENV_DEVELOPMENT' in the webpack.config need to be fixed and then we are nearly there!

Inao0:
Thanks for the scene !

I tried to reproduce the issue but could not trigger it. Someone here triggered a soft lock in one of our scenes by zooming in very quickly and then waiting for the model to upgrade. The part of the model that was on screen remained in thumb and never upgraded. Maybe it was the same issue ? I was moving through your scene really slowly because I was trying to track the quality changes.

I implemented a fix for this and lowered a bit the priority threshold without seeing any problematic stability issue.

Let me know if this version works better for you.

gjcope:
Merged to [rc-53](https://github.com/Smithsonian/dpo-voyager/tree/rc-53)

PR #325, opened 2024-11-14
use renderer's compileAsync to speed up models injection

sdumetz:
Time-to-full-load stays the same, but no loss either. In my tests it resulted in a consistent ~30% decrease of the "delayed animation frame" time when the model is first shown.


sdumetz:
Closed by https://github.com/Smithsonian/dpo-voyager/pull/332

PR #327, opened 2024-11-15
Unit tests

sdumetz:
Starting with "I'd like to write a bunch of unit tests to be sure it works as I expected", I ended up having to do a **LOT** of changes just to have the existing ~50 unit tests to run and be able to add a few more.

I'd definitely like a simpler approach that need less changes but couldn't find one.

If this ends up merged I'll have a bunch more tests on the way to make it a bit more relevant but now that I have it running I'm not 100% convinced it's worth the trouble.

Some of those changes (mainly 741e741a6a05548ee136d056052854511bf5aeab) _might_ well break something elsewhere and I didn't even bother to test it past the "it compiles" phase because that's beyond the point until we agree if this should move forward or not.

What do you think @gjcope ?


gjcope:
Generally speaking I'm all in favor of unit tests. There's just a lot to unpack here.

sdumetz:
> There's just a lot to unpack here.

Unfortunately, there is. In my experience, unit tests are a lot less useful when they require a lot of maintenance to just keep working and I suspect this is the case here. 

Just having to keep `tsc`  AND `webpack` build systems running looks like a "worse of both world"... We'll keep breaking one or the other every other day.

I'll probably come back to it at some point and try to simplify the build.

PR #328, opened 2024-11-22
fix language initialization from scene.setup

sdumetz:
remove bad check for language value. Find another way to force language from queryString

wasn't working properly since https://github.com/Smithsonian/dpo-voyager/pull/280 because of a dumb mistake of mine.

Forcing the language from `props` in `postLoadHandler` should work everytime, while allowing the value from `scene.setup.language` to be used when applicable.


gjcope:
Merged to rc-46

Issue #185, opened 2023-02-09
Memory leak when using reloadDocument()


sdumetz:
Some memory isn't released when changing documents. It's especially visible with large models as hundreds of MB go missing on each reload.

Using `WebGLRenderer.info`, I see increased textures and geometries usage over time.

Calling `WebGLRenderer.dispose()` on each document change releases the memory properly. Using this ugly hack confirmed normal memory usage for me : 
```
diff --git a/source/client/ui/SceneView.ts b/source/client/ui/SceneView.ts
index 7894cbe..9c84ced 100644
--- a/source/client/ui/SceneView.ts
+++ b/source/client/ui/SceneView.ts
@@ -117,6 +117,11 @@ export default class SceneView extends SystemView
         this.splitter.layout = EQuadViewLayout.Single;
 
         this.manipTarget.next = this.view;
+        this.system.getMainComponent(CVDocumentProvider).on("active-component", (e:any)=>{
+            if(!e.next){
+                this.view.renderer.dispose();
+            }
+        }, this);
     }
 
     protected connected()
```
calling `dispose()` on the renderer is vaguely documented over the internet but it's generally recommended when the renderer is not to be used anymore... It doesn't cause any trouble here but still feels wrong. It *might* be possible to release memory without calling `renderer.dispose()` but I couldn't find it. 

If `renderer.dispose()` is really required, there might be a good place to call it? 

I'd like some advice on this to find a better solution.

gjcope:
Hi @sdumetz - Three.js doesn't release the geometry/material/texture resources automatically due to possible shared use so that has to be done manually by calling dispose on each, though we are doing that and it seems to be working as expected. When I profile in Chrome and switch between scenes I'm scene the allocations released when the new scene is loaded. There could be other unreleased resources tied to WebGLRender, but if you are seeing hundred of MB I would assume it is model related as I am seeing nothing like that.

Do you have any example models you can share that replicate the problem so we can be looking at the same data?

gjcope:
Also FWIW I tried your code above (calling dispose on the renderer) and I'm not seeing any difference in memory released. Are you doing anything custom with model loading/rendering in your fork?

sdumetz:
Not doing anything too fancy with loading or rendering (or shouldn't be, code is not entirely mine and still a bit messy).

Platform is electron on linux, it should work like Chrome.

Models are ~1M poly each with a large texture.

[...]

PR #332, opened 2024-11-29
Model load refactor

sdumetz:
Fixes #325 and #326 with proper LoadingManager usage.

In the end it's arguably better than before because GLTFLoader counts 1 "item" per GLTF buffer, which generally doesn't make much sense and is a bit too verbose to my taste.

a581341bf5564cfbf0d4b129db2fba36978b4d60 in particular _may_ lead to some unexpected results : Those kind of algorithms are always a bit tricky to get right.

### DIFF
--- a/source/client/io/ModelReader.ts
+++ b/source/client/io/ModelReader.ts
@@ -17,11 +17,11 @@
 
 //import resolvePathname from "resolve-pathname";
 import UberPBRAdvMaterial from "client/shaders/UberPBRAdvMaterial";
-import { LoadingManager, Object3D, Scene, Mesh, MeshStandardMaterial, SRGBColorSpace } from "three";
+import { LoadingManager, Object3D, Scene, Mesh, MeshStandardMaterial, SRGBColorSpace, LoaderUtils } from "three";
 
 import {DRACOLoader} from 'three/examples/jsm/loaders/DRACOLoader.js';
 import {MeshoptDecoder} from "three/examples/jsm/libs/meshopt_decoder.module.js";
-import {GLTFLoader} from 'three/examples/jsm/loaders/GLTFLoader.js';
+import {GLTF, GLTFLoader} from 'three/examples/jsm/loaders/GLTFLoader.js';
 import { KTX2Loader } from 'three/examples/jsm/loaders/KTX2Loader.js';
 
 import UberPBRMaterial from "../shaders/UberPBRMaterial";
@@ -38,6 +38,8 @@ export default class ModelReader
     protected loadingManager: LoadingManager;
     protected renderer: CRenderer;
     protected gltfLoader :GLTFLoader;
+
+    protected loading :Record<string, {listeners : {onload: (data:ArrayBuffer)=>any, onerror: (e:Error)=>any, signal:AbortSignal}[], abortController :AbortController}> = {}
 
     protected customDracoPath = null;
 
@@ -100,24 +102,14 @@ export default class ModelReader
         return ModelReader.mimeTypes.indexOf(mimeType) >= 0;
     }
 
-    get(url: string): Promise<Object3D>
+    get(url: string, {signal}:{signal?:AbortSignal}={}): Promise<Object3D>
     {
         this.loadingManager.itemStart(url);
-        return new Promise<Object3D>((resolve, reject) => {
-            this.gltfLoader.load(url, gltf => {
-                resolve(this.createModelGroup(gltf));
-            }, null, error => {
-                if(this.gltfLoader === null || this.gltfLoader.dracoLoader === null) {
-                    // HACK to avoid errors when component is removed while loading still in progress.
-                    // Remove once Three.js supports aborting requests again.
-                    resolve(null);
-                }
-                else {
-                    console.error(`failed to load '${url}': ${error}`);
-                    reject(new Error(error as any));
-                }
-            })
-        }).then((result)=>{
+        let resourcePath = LoaderUtils.extractUrlBase( url );
+        return this.loadModel(url, {signal})
+        .then(data=>this.gltfLoader.parseAsync(data, resourcePath))
+        .then(gltf=>this.createModelGroup(gltf))
+        .then((result)=>{
             this.loadingManager.itemEnd(url);
             return result;
         }, (e)=> {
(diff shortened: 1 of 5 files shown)


gjcope:
Thanks for the PR! Looks like it's almost there. I'm not seeing any issues with regular use, but when reloading/replacing a model in place the loading spinner (and subsequently interaction prompt) disappears. This was not the case with #326 I'll try to dig in soon but feel free if you have any ideas.

source/client/io/ModelReader.ts
```diff
@@ -100,28 +102,86 @@ export default class ModelReader
...
+        this.loadingManager.itemStart(url);
+        let resourcePath = LoaderUtils.extractUrlBase( url );
+        return this.loadModel(url, {signal})
+        .then(data=>this.gltfLoader.parseAsync(data, resourcePath))
+        .then(gltf=>this.createModelGroup(gltf))
+        .then((result)=>{
+            this.loadingManager.itemEnd(url);
+            return result;
+        }, (e)=> {
+            this.loadingManager.itemError(url);
```
gjcope, line 116: Could change this to `signal.aborted ? this.loadingManager.itemEnd(url) : this.loadingManager.itemError(url);` I *think* the only ramification here is the loaded file count.

gjcope:
So turns out the issue I saw is only with a dev build due to the error thrown in the itemError handler:
https://github.com/Smithsonian/dpo-voyager/blob/5728312b12d2d6e9515aaf9a380e10e33196b562/source/client/components/CVAssetManager.ts#L208

We could just change that to a log statement or adjust the ModelReader error handler as shown in my inline comment.

source/client/io/ModelReader.ts
sdumetz: We could. Do you think the AbortError should be silenced? Maybe we should let it surface at least initially to help troubleshoot if it aborts when it shouldn't.

sdumetz: Looks like proper use is to always call itemEnd even after an error

https://github.com/mrdoob/three.js/blob/dev/src/loaders/FileLoader.js#L261

gjcope: IMHO an AbortError is an abort without a reason. But I think it's fine how it is. There is still a little wonkiness with how we are handling the interaction prompt, but I think that's a separate issue that is just highlighted by this update.

gjcope:
I think this is good to go. It will also need a node types update to package.json but I can add that in. I will also bump the "engines" node minimum to v16.

gjcope:
Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-47
#325 and #326 can be closed, yes?

gjcope:
@sdumetz We did some team testing of v0.47 today and some issues with the loading updates arose. I'd really love to push this update out, but I think at least a couple of these issues are stopping that. I'll try to list them out below as best I can. You can see some proposed solutions (or patches) here: https://github.com/Smithsonian/dpo-voyager/tree/dev-loading-fixes

High priority (these need to be resolved)
- Dragging/dropping .gltf packages is broken in standalone mode. This is due to the url transform being lost in the processing of the dependent files (.bin, textures) since the loader is using it's contextual loading manager rather than the main one. Adding the manager back to the GLTFLoader constructor fixes the issue and doesn't seem to have any other side affects other than going back to counting every buffer as an item.
- It seems pretty easy to get the loading spinner to malfunction (majorly delayed or just not visible) when rapidly reloading objects in a scene. Likely not a common use case, but users also reported seeing this in regular multi-model scenes.

Medium priority (would be nice to resolve)
- Issues have popped up when switching scenes due to independent textures (environment map, overlay map) having the models they are associated with disposed of before their loader completes. I added a couple of patches for this in the branch linked above, but they can result in those maps sometimes not getting the initial render they need (can be reproduced if you throttle your network speed). It could be good enough for now, but I think the real solution is to have a cancellable TextureLoader as well.

Low priority
- It seems to only happen when rapidly switching models, but the compileAsync function can throw an error. The dispose() code in #185 can also pretty easily trigger a weird renderer error as well but again just when rapidly switching. Curious if you are still using the #185 snippet. I can definitely see the memory leak without it and with this update.

gjcope:
> Not sure I understand what happens in the first issue. I didn't test with non-binary gltf files. I'll try to do so to double-check what happen s but in any case I'm OK with the proposed solution. Maybe in that case also add it back to the Draco/KTX loaders for consistency?
> 

When using standalone mode, there is a url modifier in place
https://github.com/Smithsonian/dpo-voyager/blob/5728312b12d2d6e9515aaf9a380e10e33196b562/source/client/components/CVStandaloneFileManager.ts#L68
to translate to the in-memory blobs. 

When loading .gltf with the additional sidecar files, those are not using our file loader instance, but another loader instance within the GLTFParser class. For that instance to also make use of the url modifier, it has to be initialized with the same file manager.

gjcope:
I was able to fix a small memory leak with an undisposed material and also identify a leak related to shadows, but neither seems to be a significant contributor.

sdumetz:
Done some research about the LoadingManager thing, it looks like having to register additional items into the loadingManager is common practice anyway (from [GLTFLoader.js#L212](https://github.com/mrdoob/three.js/blob/master/examples/jsm/loaders/GLTFLoader.js#L212)). So I was wrong to consider this number should more or less represent the asset count in the scene. I couldn't find a place where we might still be missing an `itemStart` that would be causing the instabilities though.

By the look of it, the inner code of GLTFLoader does not clean up in case an error happens halfway through parsing. We have no way of doing so though, so not handling that.

13550c16 should take care of the created buffers if the request is cancelled or if compileAsync fails.
I think `compileAsync` can be factored into the request deduplication mechanism I backported from FileLoader, to handle possible parallel calls but that would require additional testing and benchmarking so it's better left for later.

gjcope:
I think the instability ties into the CVAssetManager logic that sets busy to false when handling an error (in this case a cancelled load) 
https://github.com/Smithsonian/dpo-voyager/blob/5728312b12d2d6e9515aaf9a380e10e33196b562/source/client/components/CVAssetManager.ts#L212
The issue is pretty easy for me to reproduce and it seems like when switching scenes files can get queued up for load before the error is handled meaning busy gets set to false for the majority of the load. This still doesn't explain why it happens sometimes when *not* switching scenes.

sdumetz:
I never even noticed we used an extension of `LoadingManager` and not the base class itself. What is the reasoning behind using the `_isBusy` property instead of  `itemsLoaded === itemsTotal` (What [LoadingManager](https://github.com/mrdoob/three.js/blob/dev/src/loaders/LoadingManager.js#L49) does to trigger `onLoad`)?

Switching back to this behavior would be consistent with THREE's internal handling and might solve the problem?

gjcope:
> I never even noticed we used an extension of `LoadingManager` and not the base class itself. What is the reasoning behind using the `_isBusy` property instead of `itemsLoaded === itemsTotal` (What [LoadingManager](https://github.com/mrdoob/three.js/blob/dev/src/loaders/LoadingManager.js#L49) does to trigger `onLoad`)?
> 
> Switching back to this behavior would be consistent with THREE's internal handling and might solve the problem?

I can't speak to the reasoning because I didn't write that code, but itemsLoaded and itemsTotal are only available in the onStart and onProgress callbacks so to be to have a queryable 'isBusy' we have to maintain that state somewhere in the derived class no?

FWIW just not setting _isBusy to false when handling an error seems to do the trick and hasn't had any side affects that I can see.

sdumetz:
> FWIW just not setting _isBusy to false when handling an error seems to do the trick and hasn't had any side affects that I can see.

Yes, they are in the constructor's closure so the only access we have to this is through the `onload` callback that gets called when  `itemsLoaded === itemsTotal`, which is now the only place isBusy gets set to false.

I suppose the reasoning was so `itemEnd` can be skipped when `itemError` is called. But then we loose the counter side-effect so this is not a very good tradeof in more complex scenarios.

I don't see any reason why it would not work properly with your change, and if it does it's probably a missing `itemEnd` somewhere.

gjcope:
Good point, i missed the onload. But it seems like we are effectively doing the same thing now. _isBusy is a proxy for `itemsLoaded === itemsTotal` which we need in that class so the associated component can check it, etc. Happy to do something cleaner down the road if proposed but I think this will work for now.

PR #334, opened 2024-12-12
Prevent Missing articles. Fix #333

### LINKED ISSUE #333: users breaking the article.URI field
I had a lot of non-technical users trying to use voyager to create scenes using Voyager and an issue that came up a number of times was "disappearing articles".

Afaik 100% of them are user-errors. Some internationalization issues might compound into this but I couldn't reproduce the reported cases myself). Nevertheless, the UI should (try to) prevent this from happening at all.

What seems to happen is they edit the "Article.URI" field in `ArticleTaskView` (for unclear reasons) and can't ever find their article back.

I've [put together a POC](https://github.com/Holusion/dpo-voyager/tree/missing_articles) restricting the URI to known html assets. It works fine using an `option` property for URI selection but it's not very reliable (among other things, options list isn't dynamically linked to existing assets).

I only needed to know the article URI a few times (for manual edits where the editor was a bad fit) and never needed to manually edit the field. Are there any known cases where a user would want to edit the URI?

(note: this does not rename the file, that's done in the "media" view) 

Considering only my needs, I would make this field read-only and be done with it. Would that be ok?

sdumetz:
I added a proper error message in the article editor in case of 404 errors.

I set the `Article.URI` field to read-only.

Then I switched the  fields to use  `<sv-property-view>`because that's what it's there for.

Then I added a "textarea" mode for `<sv-property-string>`, to keep the styling roughly unchanged. This might come handy later.

Then I noticed the inconstant handling of "disabled" in `<sv-property-*>`. In particular `<sv-property-string>` relied on CSS's `pointer-events: none` for its disabled state and thus prevented easy copy-paste of the article's URI. Copying the URI if useful if someone wants to overwrite it _via_ an external PUT request. So I gathered common code in property views to hopefully limit such discrepancies in the future and added a common `disabled` property. 

I used the `aria-disabled` accessibility property which seemed semantically great for this task. Adding it as a watched property to ensure updates in the off-chance it would one day get dynamically updated.

I had trouble getting the formatting right with the property views in CVArticleTaskView and removing the absolute positioning on `ff-scroll-y` fixed it. I checked everywhere (that I knew of) it's used and it works fine with a few minor fixes.

My opinion is that this CSS positioning trick creates more problems than it solves and we should try to get rid of it everywhere we can.

It's a bit of a mess in the end, I hope it isn't too hard to read.

### DIFF
--- a/libs/ff-ui/source/styles/_styles.scss
+++ b/libs/ff-ui/source/styles/_styles.scss
@@ -168,10 +168,6 @@ $tx-medium: 0.3s;
 ////////////////////////////////////////////////////////////////////////////////
 // UTILITY CLASSES
 
-// expands the element to match the size of the parent element
-.ff-fullsize {
-  @include fullsize;
-}
 
 // displays a little downward facing triangle
 .ff-caret-down {
@@ -273,8 +269,8 @@ button, input {
 }
 
 .ff-scroll-y {
-  @include fullsize;
   overflow-y: auto;
+  max-height: 100%;
 }
 
 .ff-position-above {
--- a/source/client/components/CVArticlesTask.ts
+++ b/source/client/components/CVArticlesTask.ts
@@ -123,7 +123,7 @@ export default class CVArticlesTask extends CVTask
         if (meta && ins.create.changed) {
             const article = new Article();
             const defaultFolder = CVMediaManager.articleFolder;
-            article.uri = `${defaultFolder}/new-article-${article.id}-${ELanguageType[DEFAULT_LANGUAGE]}.html`;
+            article.uri = `${defaultFolder}/new-article-${article.id}-${DEFAULT_LANGUAGE}.html`;
 
             const standaloneFiles = this.getGraphComponent(CVStandaloneFileManager, true);
             if(standaloneFiles) {
--- a/source/client/components/CVLightTool.ts
+++ b/source/client/components/CVLightTool.ts
@@ -17,10 +17,10 @@
 
 import CLight from "@ff/scene/components/CLight";
 
-import "../ui/PropertyBoolean";
-import "../ui/PropertyOptions";
-import "../ui/PropertySlider";
-import "../ui/PropertyColor";
+import "../ui/properties/PropertyBoolean";
+import "../ui/properties/PropertyOptions";
+import "../ui/properties/PropertySlider";
+import "../ui/properties/PropertyColor";
 
 import CVDocument from "./CVDocument";
 
--- a/source/client/components/CVRenderTool.ts
+++ b/source/client/components/CVRenderTool.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import "../ui/PropertyOptions";
+import "../ui/properties/PropertyOptions";
 
 import CVDocument from "./CVDocument";
 import CVViewer from "./CVViewer";
--- a/source/client/components/CVSliceTool.ts
+++ b/source/client/components/CVSliceTool.ts
@@ -15,9 +15,9 @@
  * limitations under the License.
  */
 
-import "../ui/PropertyBoolean";
-import "../ui/PropertyOptions";
-import "../ui/PropertySlider";
+import "../ui/properties/PropertyBoolean";
+import "../ui/properties/PropertyOptions";
+import "../ui/properties/PropertySlider";
 
 import CVDocument from "./CVDocument";
 
--- a/source/client/components/CVTapeTool.ts
+++ b/source/client/components/CVTapeTool.ts
@@ -15,8 +15,8 @@
  * limitations under the License.
  */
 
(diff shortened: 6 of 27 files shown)


gjcope:
Thanks for the PR! Some inline comments below.

> I added a proper error message in the article editor in case of 404 errors.
> 
> I set the `Article.URI` field to read-only.
> 

Looks great.

> Then I switched the fields to use `<sv-property-view>`because that's what it's there for.

In this specific case yes, but there a number of other instances of this in the code like: https://github.com/Smithsonian/dpo-voyager/blob/5728312b12d2d6e9515aaf9a380e10e33196b562/source/client/ui/story/AnnotationsTaskView.ts#L132 where it serves an important purpose to allow sanitizing, enforcing restrictions, etc. 
> 
> Then I added a "textarea" mode for `<sv-property-string>`, to keep the styling roughly unchanged. This might come handy later.
> 

This seems a little inconsistent to me but I think I just need more time to consider it.

> Then I noticed the inconstant handling of "disabled" in `<sv-property-*>`. In particular `<sv-property-string>` relied on CSS's `pointer-events: none` for its disabled state and thus prevented easy copy-paste of the article's URI. Copying the URI if useful if someone wants to overwrite it _via_ an external PUT request. So I gathered common code in property views to hopefully limit such discrepancies in the future and added a common `disabled` property.
> 
> I used the `aria-disabled` accessibility property which seemed semantically great for this task. Adding it as a watched property to ensure updates in the off-chance it would one day get dynamically updated.
> 

This is all good. Love the accessibility consideration - just need to test it out a bit.

> I had trouble getting the formatting right with the property views in CVArticleTaskView and removing the absolute positioning on `ff-scroll-y` fixed it. I checked everywhere (that I knew of) it's used and it works fine with a few minor fixes.
> 
> My opinion is that this CSS positioning trick creates more problems than it solves and we should try to get rid of it everywhere we can.
> 

This seems potentially problematic but I will do some more testing.

sdumetz:
> > Then I switched the fields to use <sv-property-view>because that's what it's there for.
> 
> In this specific case yes, but there a number of other instances of this in the code [...]
 > where it serves an important purpose to allow sanitizing, enforcing restrictions, etc.

Whenever there is logic that is owned by the component we should definitely keep it this way. The sanitization is a good example of that. I only see a point in doing this if it allows removing some code, otherwise it's just pointlessly moving things around.

> > Then I added a "textarea" mode for <sv-property-string>, to keep the styling roughly unchanged. This might come handy later.
> 
> This seems a little inconsistent to me but I think I just need more time to consider it.

Not 100% sure about it either. Not a lot of place for it to be used and your example on annotations made me think maybe we ought to sanitize this too and not doing so was an oversight? If so we might have no place at all where this is used and I should just drop it.

gjcope:
> Not 100% sure about it either. Not a lot of place for it to be used and your example on annotations made me think maybe we ought to sanitize this too and not doing so was an oversight? If so we might have no place at all where this is used and I should just drop it.

The goal with sanitizing specifically was to address unsafe situations where we are allowing HTML tags to be entered by the end user and assigned to innerHTML or using unsafeHTML(). So far, that has been on a case-by-case basis as it has been requested which is admittedly inconsistent. I think it's probably fine to support at least <i> and <b> more broadly. Have you seen demand for this?

sdumetz:
> I think it's probably fine to support at least <i> and <b> more broadly. Have you seen demand for this?

Not a lot. It's just that having some "text blocks" that are "text-only" and others that are "sanitized HTML" is inconsistent. But allowing html in every text blocks isn't the only way to go about it, we could also style the HTML-enabled box differently.

sdumetz:
I left the `lead` field and the `textarea` addition to string properties aside for now.

I'll try to think of a good visual indication for "light HTML" blocks and let you know if it works.

gjcope:
Just retesting this to merge with rc-48 and noticed one other thing. The 'scroll-y' panels now have no padding at the bottom. Kind of nitpicky, but it does effect UX with panels like Tour details where there are multi-line inputs that get cut off. I played around with it a little, but can't find a solution that doesn't require adding extra elements per instance...

gjcope:
Got it, I'll give that a try. Thanks.

gjcope:
Merged to https://github.com/Smithsonian/dpo-voyager/tree/rc-48

I reverted [866550f](https://github.com/Smithsonian/dpo-voyager/pull/334/commits/866550f89245e3b795ba1b6a4f1d11486115dca3) - I think [f245c98](https://github.com/Smithsonian/dpo-voyager/pull/334/commits/f245c98e6115187e49f84bf0251d1ac0dafa140a) is fine. 

Also had to add "max-height: min-content;" to .sv-property-view to get the style right as far as I can tell. I'm not sure what style issues you were seeing before, but if you notice any others let me know. Thanks!

PR #338, opened 2025-01-10
allow document.asset.copyright to be an empty string

sdumetz:
Deleting `document.asset.copyright` in voyager-story makes for a document that fails to validate :

1. Because  `document.asset.copyright` has a `minLength`of 1 but is optional  in `document.schema.json`
2. Because CVDocument.ins.copyright.value is always serialized, even when it is an empty string, which it can be since #285 (well done, me! :wink: )

Both are reasonable, but can't work together.

At first I thought about saving `copyright` only when not empty but then Voyager will always revert back to the default Smithsonian notice on reload (and then will write it on next save), which is not ideal.

So I removed the minLength requirement and marked the property as required to reflect the behavior.

### DIFF
--- a/source/client/schema/document.ts
+++ b/source/client/schema/document.ts
@@ -53,7 +53,7 @@ export interface IDocumentAsset
 {
     type: "application/si-dpo-3d.document+json";
     version: string;
-    copyright?: string;
+    copyright: string;
     generator?: string;
 }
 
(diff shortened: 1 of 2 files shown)


gjcope:
Can it not be minLength 0 but also optional?

gjcope:
I got approval to remove the default. See if this works for your purposes: 969190e6c7c5f8bd33c7298f224abb647f72af0b

gjcope:
This should be handled in v0.48 but feel free to reopen if not.

PR #347, opened 2025-02-25
dispose of old document only if it exist

sdumetz:
https://github.com/Smithsonian/dpo-voyager/commit/af6cb9f49c92d28e6fecfb46eb4494d75f3dd013 makes `reloadDocument()` crash the application if called on an uninitialized stack.

The behavior of `reloadDocument` when there is no document to reload is not documented, so it's not strictly speaking a bug.

I'd say `reloadDocument` should check by itself if there is a document to dispose of because :

- that was the original behavior
- The user has no (easy) way to know if there _IS_ a document, `ExplorerApplication.documentProvider` being protected. (one _could_ call `ExplorerApplication.system.getMainComponent(CVDocumentProvider).activeComponent`, but we probably don't want that)

### DIFF
--- a/source/client/applications/ExplorerApplication.ts
+++ b/source/client/applications/ExplorerApplication.ts
@@ -245,7 +245,7 @@ Version: ${ENV_VERSION}
         const oldDocument = this.documentProvider.activeComponent;
         this.documentProvider.createDocument(documentTemplate as any);
         this.evaluateProps();
-        oldDocument.dispose();
+        oldDocument?.dispose();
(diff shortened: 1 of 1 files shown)


gjcope:
Thanks! Yeah the original issue was more of an order of operations problem so good catch. I'm getting a security error for some reason when trying to pull this branch so I will probably manually make this update, but it will be included in rc-49.

gjcope:
Manually merged in rc-49

PR #354, opened 2025-04-15
Annotations translation for smithsonian

Inao0:
These changes aim at making it easier to translate annotations, especially if English is not the default language of the scene.

The following improvements are included:
- When working in the default language of a scene, the table shows only one column instead of the same one twice 
image
- When working in a secondary language, show the default language (here Spanish) on the left, instead of only English
image
- When working in a secondary language, show the default language lead above the lead field
image

This preserves the behavior of always having at least a title in the default language (not necessarily English)

### DIFF
--- a/assets/language/string.resources.fr.json
+++ b/assets/language/string.resources.fr.json
@@ -41,6 +41,8 @@
   "Material": "Texture",
   "Measure": "Mesure",
   "Measured Distance": "Distance mesurée",
+  "New Annotation": "Nouvelle Annotation",
+  "New Article": "Nouvel Article",
   "No tour selected": "Pas de visite sélectionnée",
   "No tour steps defined": "Pas d'étapes de visites définies",
   "Normals": "Normales",
--- a/source/client/components/CVAnnotationsTask.ts
+++ b/source/client/components/CVAnnotationsTask.ts
@@ -258,6 +258,8 @@ export default class CVAnnotationsTask extends CVTask
 
     protected createAnnotation(position: number[], direction: number[])
     {
+        const languageManager = this.activeDocument.setup.language;
+        const language = languageManager.ins.language.value;
         const annotations = this.activeAnnotations;
         if (!annotations) {
             return;
@@ -274,6 +276,14 @@ export default class CVAnnotationsTask extends CVTask
         const model = annotations.getComponent(CVModel2);
         const annotation = new Annotation(template);
 
+        annotation.language = language;
+        annotation.title = languageManager.getLocalizedString("New Annotation");
+
+        if (languageManager.ins.language.value != ELanguageType[languageManager.defaultLanguage]){
+            //Ensure that there is a title in the default language of the scene
+            const defaultLocalized = languageManager.getLocalizedStringInDefaultLanguage("New Annotation");
+            annotation.titleInLanguage(defaultLocalized, ELanguageType[languageManager.defaultLanguage]);
+        }
         const data = annotation.data;
         data.position = position;
         data.direction = direction;
--- a/source/client/components/CVLanguageManager.ts
+++ b/source/client/components/CVLanguageManager.ts
@@ -40,6 +40,8 @@ export default class CVLanguageManager extends Component
     static readonly text: string = "Language";
     static readonly icon: string = "";
 
+    private _sceneDefaultLanguage:TLanguageType;
+    private _defaultLanguageTranslations: ITranslation = {};
     private _activeLanguages: {[key in TLanguageType]?: ILanguageOption} = {};
     private _translations: ITranslation = {};
 
@@ -73,6 +75,9 @@ export default class CVLanguageManager extends Component
     }
     get activeLanguages() {
         return Object.values(this._activeLanguages);
+    }
+    get defaultLanguage(){
+        return this._sceneDefaultLanguage;
     }
 
     /**
@@ -123,7 +128,9 @@ export default class CVLanguageManager extends Component
         data = data || {} as ILanguage;
 
         const language = ELanguageType[data.language || "EN"] ?? ELanguageType[DEFAULT_LANGUAGE];
-
+        this._sceneDefaultLanguage = data.language || "EN";
+        this.assetReader.getSystemJSON("language/string.resources." + this._sceneDefaultLanguage.toLowerCase() + ".json").then(json => 
+            {this._defaultLanguageTranslations = json});
         ins.language.setValue(language);
     }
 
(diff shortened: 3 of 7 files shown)


gjcope:
Hi @Inao0 - thanks for the PR! Some comments below.

> These changes aim at making it easier to translate annotations, especially if English is not the default language of the scene.
> 
> The following improvements are included:
> 
> * When working in the default language of a scene, the table shows only one column instead of the same one twice

I realize the current display is redundant, but have you seen complaints or usability issues due to this? I'm not sure all of the extra code needed is a good justification for the change. It also allows the interface to stay static, so it's more obvious what is going on for the user when changing languages. But totally open to further discussion if issues have come up.

> * When working in a secondary language, show the default language (here Spanish) on the left, instead of only English

I'm not sure what the change is here. If I change the Story default language (DEFAULT_LANGUAGE) in the current release it shows up on the left instead of English. I did notice that the Explorer language button doesn't update properly which is definitely a bug. Am I missing the intent here?

> * When working in a secondary language, show the default language lead above the lead field

This is great! Could we add an option to collapse/hide it? It seems like this would be super valuable to people translating in real time, but that's not how our translation workflow goes so I think it would be a little obtrusive to our users (since they can get long) if not hideable. But it's a great feature!

Inao0:
Hi @gjcope, thanks for the feedback!

> - When working in the default language of a scene, the table shows only one column instead of the same one twice

I agree its a bit more code, but having two columns with the same language seems more confusing to me.
This also allows longer titles to be displayed better on smaller screens.
I could try to push the generation of the title of the table into AnnotationList.ts to reduce the amount of if/else and clean AnnotationsTaskView.ts a little bit more.

>  When working in a secondary language, show the default language (here Spanish) on the left, instead of only English

Here I am talking about the default language of the scene instead of the DEFAULT_LANGUAGE. Changing DEFAULT_LANGUAGE requires re-building and re-deploying the application.
Having the language on the left being the scene's default language gives more flexibility. This allows one to work one time on a scene primarily in English and one time on a scene primarily in French using the same application.

> - When working in a secondary language, show the default language lead above the lead field
This is great! Could we add an option to collapse/hide it?

It is already hidden when working in the scene's default language.
Should this be an application wide option, to disable it completely?
I could put an option in /source/client/schema/common.ts next to DEFAULT_LANGUAGE.

gjcope:
> I agree its a bit more code, but having two columns with the same language seems more confusing to me. This also allows longer titles to be displayed better on smaller screens. I could try to push the generation of the title of the table into AnnotationList.ts to reduce the amount of if/else and clean AnnotationsTaskView.ts a little bit more.

I'm still not seeing the value for the additional complexity but it's also tied in a little bit with the below issue, so maybe if that is resolved this will be less of a concern.

> Here I am talking about the default language of the scene instead of the DEFAULT_LANGUAGE. Changing DEFAULT_LANGUAGE requires re-building and re-deploying the application. Having the language on the left being the scene's default language gives more flexibility. This allows one to work one time on a scene primarily in English and one time on a scene primarily in French using the same application.

Currently, the language default for the Explorer scene is meant to be whatever has been 'captured' in the default scene state and ties to the intended audience for the scene. The Story default language ties to the preferred language of the scene (or content) creator. Eventually, when we have the translations available, the intent is for the Story UI strings to change according to the Story default as well. I'm not opposed to this changing, just concerned that by making things simpler for this use case we may be making it more complicated for others.

For instance, I could be crafting a scene for a primarily Spanish-language site but be more comfortable with English as my own default. Any time I opened that scene to make an edit, I would expect the authoring interface to still be in my own preferred language.

Maybe the Story default language needs to be more easily configured?

Also, I think there are a number of other things missing for the proposed implementation. If the 'content default' language is now the scene default instead of Story default, it would need to change (and update the UI) whenever the scene default changes. Right now it looks like it only happens on load. This is also not addressing the other multilingual content tasks (Articles and Tours). We also need to take into account the many users we have that use Story in 'standalone' mode, meaning no scene data is initially loaded.

Apologies for the long response but I want to make sure we get this right before redefining something fundamental. Hopefully we can land on something that improves things for everyone!

>It is already hidden when working in the scene's default language. Should this be an application wide option, to disable it completely? I could put an option in /source/client/schema/common.ts next to DEFAULT_LANGUAGE.

No, I think just an option in the UI, like making it a collapsible section would work. I showed it to content folks on my side and they loved it. They just felt it would be more usable if there was a way to hide it.

Inao0:
Hello @gjcope,

Regarding the lead translation, I can make it collapsible and submit a pull request with just that. It might be easier than mixing it with the "how to handle languages" part.

I have discussed your feedback with @sdumetz. Just to clarify, this pull request was only about annotations so that we could have some something to discuss. I was planning to work on Articles and other parts of the interface next once we have something we can agree on.

We propose that I work on introducing/reorganizing languages so that we have in the language manager : 

- A **UI language** : for all the interface tools and buttons. Ideally we introduce a voyager story setting somewhere to modify it. It prevents menus from changing as you change your current language.

- A **scene setup language**  : This is the "primary" language of the scene. It can be different from the UI language, your UI language may not even be available in the scene (see below where my UI is in English, but I am working on a scene in Spanish and French). This is the one that always has at least a default value for its fields, no matter what. It is always shown on the left. We can keep two columns at all time if you feel it is better. 
image

> Also, I think there are a number of other things missing for the proposed implementation. If the 'content default' language is now the scene default instead of Story default, it would need to change (and update the UI) whenever the scene default changes. Right now it looks like it only happens on load. This is also not addressing the other multilingual content tasks (Articles and Tours). We also need to take into account the many users we have that use Story in 'standalone' mode, meaning no scene data is initially loaded.

By default, the scene setup language could be set to the UI language for people using Story in 'standalone' mode.
Yes, changing the default language should update the UI, but only in the parts relevant to content creation. This is definitely missing from the current implementation, thank you for pointing that out.

- A list of **available languages** : all the languages already available in the scene. No changes here.

- A **current language** : The language you are currently working on. Shown in the explorer. Shown in all UI used for content creation and translation.

Would this work ?

gjcope:
> Regarding the lead translation, I can make it collapsible and submit a pull request with just that. It might be easier than mixing it with the "how to handle languages" part.

Sounds good.
> 
> I have discussed your feedback with @sdumetz. Just to clarify, this pull request was only about annotations so that we could have some something to discuss. I was planning to work on Articles and other parts of the interface next once we have something we can agree on.

Sure, that makes sense. We will just need the full implementation before being able to merge.

> 
> We propose that I work on introducing/reorganizing languages so that we have in the language manager :
> 
> * A **UI language** : for all the interface tools and buttons. Ideally we introduce a voyager story setting somewhere to modify it. It prevents menus from changing as you change your current language.
> * A **scene setup language**  : This is the "primary" language of the scene. It can be different from the UI language, your UI language may not even be available in the scene (see below where my UI is in English, but I am working on a scene in Spanish and French). This is the one that always has at least a default value for its fields, no matter what. It is always shown on the left. We can keep two columns at all time if you feel it is better.

This works, and I can see how it well be beneficial for your use cases. I think we should also add in some more explicit labeling to communicate this since it is a departure from the current convention. The current one could use it as well, but this would be a good opportunity to improve usability here.

> 
> By default, the scene setup language could be set to the UI language for people using Story in 'standalone' mode. Yes, changing the default language should update the UI, but only in the parts relevant to content creation. This is definitely missing from the current implementation, thank you for pointing that out.
> 
> * A list of **available languages** : all the languages already available in the scene. No changes here.
> * A **current language** : The language you are currently working on. Shown in the explorer. Shown in all UI used for content creation and translation.
> 
> Would this work ?

Sounds good, thanks!

Inao0:
Hello @gjcope,
I am working on the new version of the language manager and I have a few questions about the behavior of annotations.
I discovered that when an annotation is selected and you create a new one, it copies the selected one.
From what I understood, it is maybe useful for keeping the styling and type of annotation ?
Why copy the title and the lead ?
At first I was really confused by this and thought It was a bug.
Is there a reason to keep it that way ? What do you think about default values instead like "New annotation" title and empty lead ?

gjcope:
Hi @Inao0 , is this this PR still wanted? I think some of this was addressed in subsequent PRs, but not sure. Thanks!

PR #357, opened 2025-05-15
Add a decomposition option to the measuring tape tool

Inao0:
This additional feature adds a decomposition option to the measuring tape tool, allowing one to measure along the 3 axes of the scene/world.

For example, it makes it possible to measure the height of an object when the highest point is not above the lowest point (see picture below).

The measures taken along each axis are shown on the model as thin colored lines. Their values are displayed in the toolbox.

This tool is currently intended more for people exploring a model, rather than for the tours.
No additional tags displaying measure values are shown to limit visual noise. The decomposition is disabled by default.

image


Inao0:
Sorry for the noise, I forgot to checkout the Smithsonian branch. Will redo a cleaner pull request.

PR #358, opened 2025-05-15
 Add a decomposition option to the measuring tape tool #357 

Inao0:
This additional feature adds a decomposition option to the measuring tape tool, allowing one to measure along the 3 axes of the scene/world.

For example, it makes it possible to measure the height of an object when the highest point is not above the lowest point (see picture below).

The measures taken along each axis are shown on the model as thin colored lines. Their values are displayed in the toolbox.

This tool is currently intended more for people exploring a model, rather than for the tours.
No additional tags displaying measure values are shown to limit visual noise. The decomposition is disabled by default.
image


gjcope:
Hey @Inao0 , thanks for the bump. No issues jump out at me with the implementation - nice work!. I will just need to run it by my team to weigh in on the value/UI complexity trade off. It's in the queue!

gjcope:
Hi @Inao0 ,
This was reviewed by my team and the consensus is that it is a good value add, but could use a few modifications:

gjcope:
Hey @Inao0 , thanks for the updates - 

> No problems with hiding the labels. The idea behind showing them was to avoid buttons moving around when the option is toggled. I did not consider the amount of space they are taking on mobile screen.
> 
> How about "along axes"? I would like to find a way to clearly indicate that the second button is an option for the tape tool. I think it might look like another independent tool if it just says "Axes". I could also show it only if the tape tool is activated, but it will lead to a lot of "jumps" in element positions.
> 

I don't think the 'jump' is necessarily a bad thing as it calls attention to the UI, but agree multiple moves would be excessive. I'm not sure 'along axes' is that meaningfully different from 'axes', but I will run it by some users on our end and see what they think.


> > 2. Responsive layout - I don't think there is a way around this flowing to two lines when on mobile, but it would be good if we could keep the new functionality grouped on the second line. Right now it is highly likely to split. See pic. Since the units will be the same for each measurement, maybe we can incorporate it once in a more efficient way?
> 
> I will look at keeping them grouped.
> 
> I tried showing the units once at the end of the line, but this looks weird to me : 
> However, if we remove "axes" from the labels, we can show the units there.

I was thinking something like Axes (m), but I think adding to the label is a good compromise. The main goal is not stretching out the horizontal real estate with redundant info which I think this accomplishes.

> 
> > 3. Some people found the axis measurement lines difficult see. I'm not sure if the solution here, but maybe we could experiment with brighter hues?
> 

Looks pretty good in the picture, but we should really user test it to be sure.

PR #359, opened 2025-05-21
Separating UI Language from scene primary language and active language

Inao0:
Following the discussion in #354, here is the new proposal for the Language Manager with :
- *activeLanguage* (ins/outs) : Current language of the scene
- *uiLanguage* (ins/outs) : Language of the UI. Initialized from `DEFAULT_LANGUAGE`.
- *sceneSetupLanguage* : The primary language of the scene.
- *_sceneLanguages* : All languages available in the scene.

The UI language can be set using the a dropdown menu in the top left corner. The left columns of tables in Annotations, Articles, ...  always correspond to the scene setup language. If the scene state is captured, the scene setup language is updated. 

In standalone mode, the scene setup language is initialized with `DEFAULT_LANGUAGE.` 

Where possible, hard-coded strings were replaced in order to use the language manager.

In the following image, the UI language is French, the scene setup language is Spanish and the active language is English.
 
image

**Limits**
This pull request does not include the translation of : 
- Properties views
- Some hard-coded column names in the notes / tours tables in the bottom panels
- The article editor
- Default values in JSON objects for creating new annotations, articles ...

**Something to check :**
In `source/client/ui/explorer/NodeView.ts` I commented out a line to prevent errors when disconnecting. Calling onActiveNode() triggers a UI refresh while the active document is null. This was raising errors when trying to access the language manager. This seemed to work fine like that but let me know if there is an issue with that that I did not catch.

``` Typescript
    protected disconnected()
    {
        const provider = this.nodeProvider;
        provider.off<IActiveNodeEvent>("active-node", this.onActiveNodeEvent, this);

        if (provider.activeNode) {
            this.activeNode = null;
           //this.onActiveNode(provider.activeNode, null);  HERE
        }

        super.disconnected();
    }
    
    protected onActiveNode(previous: NVNode, next: NVNode)
    {
        this.requestUpdate();
    }

```


gjcope:
Thanks for the new PR @Inao0 !

> Following the discussion in #354, here is the new proposal for the Language Manager with :
> 
> * _activeLanguage_ (ins/outs) : Current language of the scene
> * _uiLanguage_ (ins/outs) : Language of the UI. Initialized from `DEFAULT_LANGUAGE`.
> * _sceneSetupLanguage_ : The primary language of the scene.
> * __sceneLanguages_ : All languages available in the scene.

I would suggest we change 'sceneSetupLanguage' to something like primarySceneLanguage or defaultSceneLanguage to make the differentiation a little clearer.

> 
> The UI language can be set using the a dropdown menu in the top left corner. The left columns of tables in Annotations, Articles, ... always correspond to the scene setup language. If the scene state is captured, the scene setup language is updated.
> 

On the backend this all seems to work great but I think we need a different solution for the UI. The top bar is valuable real estate and the dropdown takes up quite a bit of space for functionality that I don't envision will be used beyond the initial opening of the app. Maybe it could be set via a url parameter or moved to a config tab? Open to suggestions.

> Where possible, hard-coded strings were replaced in order to use the language manager.
> 

Thanks for getting a start on this! We probably won't promote this until we are closer to full support to avoid confusion, but I appreciate the effort. It also makes me think about how we will handle 'supported' languages for the UI. Right now, we don't release a new content language until its dictionary is complete (at least at the time of publishing). Once the backend support is complete, it's likely we still won't have UI-complete dictionaries for the supported languages for some time...  

> 
> **Something to check :** In `source/client/ui/explorer/NodeView.ts` I commented out a line to prevent errors when disconnecting. Calling onActiveNode() triggers a UI refresh while the active document is null. This was raising errors when trying to access the language manager. This seemed to work fine like that but let me know if there is an issue with that that I did not catch.

Do you have a small example on how to reproduce this issue? I'd like to take a closer look.

A couple other notes:
- Keep an eye out for consistent usage of the language getter vs the private variable. Also the use of ELanguageType[this.language] so that the content arrays are correctly indexed.
- As mentioned in https://github.com/Smithsonian/dpo-voyager/pull/354#issuecomment-2838987742 I think we should change the content column labels to make this update more obvious. Maybe something like "Default (language_code)" and "Active (language_code)"? Not married to that, but think more explicit labeling could be helpful.

Inao0:
Hello @gjcope,

I moved the UI language control to an url / html element parameter. If one day there are more useful parameters for Voyager UI, a control panel can be created at that time.

> - Keep an eye out for consistent usage of the language getter vs the private variable. Also the use of ELanguageType[this.language] so that the content arrays are correctly indexed.
> - As mentioned in https://github.com/Smithsonian/dpo-voyager/pull/354#issuecomment-2838987742 I think we should change the content column labels to make this update more obvious. Maybe something like "Default (language_code)" and "Active (language_code)"? Not married to that, but think more explicit labeling could be helpful.

The new commits should address these points as well as renaming `sceneSetupLanguage` to `primarySceneLanguage`. Please let me know if I have missed anything else.

> Do you have a small example on how to reproduce this issue? I'd like to take a closer look.

You can remove the comment at line 57 in `NodeView.ts` and try switching between the different task views (Pose, Capture, Derivatives, Annotations, etc.) 
When leaving a view, the corresponding errors appear in the console :
```
Uncaught (in promise) TypeError: this.activeDocument is null
    render CaptureTaskView.ts:48
    update lit-element.ts:300
    performUpdate updating-element.ts:775
    _enqueueUpdate updating-element.ts:725
    requestUpdateInternal updating-element.ts:690
    requestUpdate updating-element.ts:708
    onActiveNode NodeView.ts:65
    disconnected NodeView.ts:57
    disconnected CVTask.ts:188
    disconnected CaptureTaskView.ts:40
    disconnectedCallback CustomElement.ts:175
    removeNodes dom.ts:52
    clear parts.ts:356
    __commitNode parts.ts:263
    commit parts.ts:243
    update template-instance.ts:49
    __commitTemplateResult parts.ts:291
    commit parts.ts:241
    render render.ts:51
    render shady-render.ts:277
    update lit-element.ts:305
    performUpdate updating-element.ts:775
    _enqueueUpdate updating-element.ts:725
    requestUpdateInternal updating-element.ts:690
    requestUpdate updating-element.ts:708
    onUpdate CustomElement.ts:197
    emit Publisher.ts:221
    set activeComponent CComponentProvider.ts:159
    onClickTask TaskBar.ts:109
    handleEvent parts.ts:532
    __boundHandleEvent parts.ts:490
    commit parts.ts:523
    update template-instance.ts:49
    __commitTemplateResult parts.ts:300
    commit parts.ts:241
    render render.ts:51
    render shady-render.ts:277
    update lit-element.ts:305
    performUpdate updating-element.ts:775
    _enqueueUpdate updating-element.ts:725
    requestUpdateInternal updating-element.ts:690
    initialize updating-element.ts:531
    initialize lit-element.ts:224
    UpdatingElement updating-element.ts:516
    LitElement lit-element.ts:101
    CustomElement CustomElement.ts:35
    SystemView SystemView.ts:29
    TaskBar TaskBar.ts:41
    firstConnected MainView.ts:168
    connectedCallback CustomElement.ts:164
    customElement CustomElement.ts:205
    __decorate MainView.ts:43
    <anonymous> MainView.ts:81
    ts voyager-story.js:6164
    __webpack_require__ voyager-story.js:10616
    <anonymous> voyager-story.js:10692
    <anonymous> voyager-story.js:10694
CaptureTaskView.ts:48:32
```

Inao0:
Hello @gjcope, thanks for the feedback !

> That sounds good. I think this implementation has an issue though. It loads up the UI in the correct language when setting the 'uiLang' parameter, but clicking on any of the task buttons reverts it back to default.

It should be fixed now. If not, can you give me more details to reproduce the issue ? 

I also worked on the behavior of the standalone mode : if no `lang` parameter is specified, it uses the `uiLang` parameter (if present) to setup the language of the scene. It seems more intuitive than keeping `DEFAULT_LANGUAGE` when a `uiLang` is specified.

The indexing issue you mentioned in your comment has been fixed in the last commit.

> We definitely need that line or we will have a memory leak from event handlers not being removed from various views when disconnected. I believe this has come up in the past as we are explicitly checking for activeDocument in the annotation and article view render functions. It could be handled this same way, or onActiveDocument could be overridden to set the language.

I copied the activeDocument check from the `render()` method to the other TaskViews that needed it. I put back the line I previously removed and it is now working fine (no more errors in the console when switching tasks).

> I also noticed the 'undefined' values now appear as blank in the content columns. Was this intentional? I welcome input on it as you all are more frequent multilingual authors, but I find it to be more helpful and also is consistent with the 'undefined' that still shows up in the task details.

It was not intentional. I opted for the default English text for now, instead of "undefined", which is not very friendly looking. Ideally, I would like to have language-specific placeholders in the long run. This is probably outside the current scope of this pull request.

gjcope:
> 
> If not "New annotation" and its variants, what about "Missing translation" ?

I talked it over with our content folks and they agreed that "New annotation", etc. was not optimal but liked your "Missing translation" suggestion. Though it seems like that might be a little long for the UI space we have, so maybe just "Missing"? Or possibly "MISSING" just to make sure it jumps out to users?

Inao0:
This implements the red "Missing translation" placeholder in the Annotations, Articles, Tours and Tour editor lists.

> The only inconsistency I see is that the tour menu in Explorer still shows 'undefined' for undefined language entries.

This has been fixed in the last commit. In the Explorer's tour menu, the lead shows "Missing translation" only if there is at least one nonempty lead for the tour, in any language.

gjcope:
@Inao0 There was some confusion during user testing with the "Missing translation" text as it also shows in various places when working with a single language scene. We've changed it to "Missing content" to try and address that in [rc-53](https://github.com/Smithsonian/dpo-voyager/tree/rc-53). Please let me know if you have any concerns.

PR #369, opened 2025-08-01
Update build system

sdumetz:
I did a bunch of cleanup while trying to improve our internal build process :

1. removed the hard-coded node_module and three path : it is now entirely unnecessary and removing them allows the default resolve behavior to apply: less surprises.
2. I removed the git tag check that prevented building from anything but a git repository (eg: a zip archive or a tagless clone). This **should** be equivalent since tags are synchronized with releases but it may not be depending on your internal processes. if it's the case I'm OK  having a `catch` fallback so this only happens in the mentioned cases.
3. I got into trouble again with `sass` so I ran [sass-migrator](https://github.com/sass/migrator#readme) to ensure `dart-sass` 2.x compatibility and make the deprecation messages go away.
4. I removed raw-loader that annoyed me with its deprecation messages. Replaced with `asset/source`. Only visible difference is it exports es6 modules so I had to rewrite the `*(frag|vert)` requires.
5. Apparently `some_condition && ENV_DEVELOPMENT` made webpack's file cache mad (for some reason) while `ENV_DEVELOPMENT && some_condition` was OK.


source/client/global.d.ts
```diff
@@ -1,10 +1,13 @@
-declare module '!raw-loader!*' {
+declare module '*?raw' {
     const contents: string;
     export = contents;
 }
 
 declare module "*.scss";
 declare module "*.css";
+declare module "*.css";
```
gjcope, line 8: Redundant declaration?

gjcope:
> I did a bunch of cleanup while trying to improve our internal build process :

Looks good and loving the lower package size as a result!

> 2. I removed the git tag check that prevented building from anything but a git repository (eg: a zip archive or a tagless clone). This **should** be equivalent since tags are synchronized with releases but it may not be depending on your internal processes. if it's the case I'm OK  having a `catch` fallback so this only happens in the mentioned cases.

Agree this **should** be fine. My only concern is that in reality sometimes user error (me 🙃 ) the package.json version doesn't get updated. Admittedly a process issue on our end, but a 'catch' could be helpful.

> 3. I got into trouble again with `sass` so I ran [sass-migrator](https://github.com/sass/migrator#readme) to ensure `dart-sass` 2.x compatibility and make the deprecation messages go away.

No issues here, but curious what the deprecation messages were that you were getting. I haven't seen them, but I'm sure it's an environment difference.

gjcope:
No worries. Thanks for the fix! Merged to rc-55.

Issue #190, opened 2023-02-22
Feature Request: Ability to alternate scene light 'type' 


campbell-ja:
**Feature Request** 

Currently only 'directional' lights seem to be possible. 

Would be great to be able to change the light _type_ to 'ambient', 'point', 'hemisphere', or 'spot'. 

thanks!

carschno:
On second thought, the challenge seems to be that changing the type of an existing light requires replacing the current light object with a new one.
Experimentally, I have approached this in the `CLight#update(context)` method:

```typescript
export default class CLight extends CObject3D
{
    [...]
    update(context)
    {      
            if (ins.type.changed) {
                // TODO: how to replace `this` with a new Light object?
            }
        [...]
    }
```

gjcope:
I think the key difference is whether we expect light management (at this level) to be supported by Explorer and Story or just Story. If this is Story-only functionality, which I imagine it is, it likely makes the most sense to take the Task or Menu route to keep unneeded code out of Explorer.

carschno:
FYI: https://github.com/Smithsonian/dpo-voyager/pull/374 adds options to add and delete light, therefore effectively solving this issue.

We have discussed the option of changing the type of an existing light, but it turned out that there are too many more or less subtle differences between the light types so that a direct type change would require a pretty hacky and unintuitive solution.

gjcope:
@campbell-ja This should now be possible through the light add/remove support added to rc-57. Actually changing "type" turned out to be a tricky thing. If this doesn't address your need, feel free to reopen!

PR #374, opened 2025-08-15
Make lights editable in UI

carschno:
~~A new Task to~~ Allow lights to be added, removed and edited by type in Story.

## Tasks
- [x] Keyboard accessibility popup menu to create lights, transfer focus to popup
- [x] Keyboard/focus for new light deletion confirmation pop-up
- [x] implement stylesheets for new div classes
- [x] UI: integrate new options for light creation, deletion and type change with other light settings
- [x] convert light intensity ~~between 0-1 range and Candela scale to match what new light type uses~~
- ~~TBD: compute default property values when creating new lights (including for changing light type for properties that do not exist on old type)~~

## Screenshot showing changes
Screenshot 2025-10-08 at 11-27-51 Voyager Story v0 56 1-40-gcd94829e DEV

### DIFF
--- a/source/client/components/CVLightTool.ts
+++ b/source/client/components/CVLightTool.ts
@@ -16,7 +16,6 @@
  */
 
 import CLight from "@ff/scene/components/CLight";
-import { INodeChangeEvent } from "@ff/graph/Node";
 
 import "../ui/properties/PropertyBoolean";
 import "../ui/properties/PropertyOptions";
@@ -27,6 +26,7 @@ import CVDocument from "./CVDocument";
 
 import CVTool, { types, customElement, html, ToolView } from "./CVTool";
 import CVEnvironmentLight from "./lights/CVEnvironmentLight";
+import NVNode from "client/nodes/NVNode";
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -63,42 +63,11 @@ export default class CVLightTool extends CVTool
 
     protected onActiveDocument(previous: CVDocument, next: CVDocument)
     {
-        this.detachLightListeners();
-
         this.lights = next ? next.getInnerComponents(CLight).filter((light) => light.ins.enabled.value) : [];
-
-        this.attachLightListeners();
-        this.refreshLightOptions();
-
+        this.ins.light.setOptions(this.lights.map(light => light.ins.name.value));
         this.outs.light.setValue(this.lights[0] ?? null);
 
         super.onActiveDocument(previous, next);
-    }
-
-    protected refreshLightOptions()
-    {
-        this.ins.light.setOptions(this.lights.map(light => light.node.name));
-    }
-
-    protected attachLightListeners()
-    {
-        this.lights.forEach(light => light.node?.on("change", this.onLightNodeChange, this));
-    }
-
-    protected detachLightListeners()
-    {
-        if (!this.lights) {
-            return;
-        }
-
-        this.lights.forEach(light => light.node?.off("change", this.onLightNodeChange, this));
-    }
-
-    protected onLightNodeChange(event: INodeChangeEvent)
-    {
-        if (!event || event.what === "name") {
-            this.refreshLightOptions();
-        }
     }
 }
 
@@ -170,6 +139,27 @@ export class LightToolView extends ToolView<CVLightTool>
         }
 
         this.requestUpdate();
+    }
+
+    protected onActiveNode(previous: NVNode, next: NVNode)
+    {
+        if (previous && previous.light) {
+            previous.light.ins.name.off("value", this.refreshLightList, this);
+            previous.light.ins.enabled.off("value", this.refreshLights, this);
+        }
+        if (next && next.light) {
+            next.light.ins.enabled.on("value", this.refreshLights, this);
+            next.light.ins.name.on("value", this.refreshLightList, this);
+        }
+    }
+
+    protected refreshLights() {
+        this.tool.lights = this.activeDocument.getInnerComponents(CLight).filter((light) => light.ins.enabled.value);
+        this.refreshLightList();
+    }
+
+    protected refreshLightList() {
+        this.tool.ins.light.setOptions(this.tool.lights.map(light => light.ins.name.value)); 
     }
 
     protected async setFocus()


source/client/components/CVLightsTask.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    update() {
+        const { ins } = this;
+
+        if (ins.create.changed) {
+            console.log("create light", ins.type.value);
+            const lightType = lightTypes.find(type => type.type === ELightType[ins.type.value]);
+            const light = this.createComponent<ICVLight>(lightType)//.fromDocument(document,  this.system.getComponent("CLight").parentComponent.parent.node );
+            // FIXME: how to add to document/parent node?
```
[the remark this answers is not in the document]
gjcope: You can see one option in this WIP branch for adding an environment light component. https://github.com/Smithsonian/dpo-voyager/blob/f2046dd84c44afebf5b0cc92d682d30787384db9/source/client/components/CVEnvironment.ts#L195 When you call createComponent() on a node, the component gets added to that node. So we create a new transform node, create the component on that, and then attach the node to the light group.

libs/ff-scene/source/components/CLight.ts
```diff
@@ -25,6 +25,7 @@ export default class CLight extends CObject3D
     static readonly typeName: string = "CLight";
 
     protected static readonly lightIns = {
+        id: types.String("Light.Id", ""),
```
carschno, line 28: Is this needed?

source/client/components/CVLightsTask.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    protected static copyAllProperties(source: CLight, target: NVNode) {
+        CVLightsTask.copyProperties((source as any).settingProperties, target.light.ins.properties);
+        if (source instanceof CDirectionalLight) {
+            // CDirectionalLight multiplies intensity by PI, so it needs to be compensated (divide by PI)
+            target.light.ins.setValues({ "intensity": source.light.intensity / Math.PI });
+        }
+
+        CVLightsTask.copyProperties((source.transform as CVNode).settingProperties, target.transform.ins.properties);
+        // TODO: copy shadow properties?
```
carschno, line 115: Not sure if shadows are part of the light, TBD.

carschno: Shadow parameters are apparently copied implicitly, as far as applicable to the new light type.

carschno:
The UI is now a bit messy with the settings per light in the `Settings` task, whereas creating, removing and editing light types happen in the newly introduced `Lights` task. However, that might be something to fix in the general UX overhaul. Otherwise, I am happy to adapt accordingly.
@gjcope Generally, I would also be happy about a slightly nitpicky review that helps me to better understand if the changes are in line with the best practices of the project for future work.

source/client/components/CVLightsTask.ts
```diff
@@ -0,0 +1,140 @@
...
+                this.nodeProvider.activeNode = newNode;
+                return true;
+            }
+        }
+        return false;
+    }
+
+    protected createLightNode(newType: ELightType, newName: string): NVNode {
+        const lightType = lightTypes.find(lt => lt.type === ELightType[newType].toString());
+
```
ctwhome, line 110: It could handle errors in case there are no lights with:
```
if (!lightType) {
      throw new Error(`Unsupported light type: ${newType}`);
  }
  ```

**Review by ctwhome — COMMENTED** [review]
- Consider adding JSDoc comments for the main public methods, especially createLightNode() and copyLightProperties(), to help future contributors understand the light creation workflow.
- The default light intensities (e.g., 50 for rect lights) could be extracted as named constants for easier configuration.
-  As you mentioned, light settings are spread across tasks. Consider adding a comment or TODO for future consolidation of light-related UI into a single, cohesive panel.

**Comment by carschno** [comment]
Closes https://github.com/Smithsonian/dpo-voyager/issues/190

File: source/client/components/CVLightsTask.ts [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+        // Copy light properties
+        (sourceLight as any).settingProperties
+            .forEach((sourceProp: Property) => {
+                targetLight.ins.properties
+                    .find(targetProp => targetProp.key === sourceProp.key)
+                    ?.setValue(sourceProp.value);
+            }
+            );
+
+        // CDirectionalLight multiplies intensity by PI, so it needs to be compensated (divide by PI)
```
gjcope, line 129: Shadow properties should also be copied (where appropriate) and/or recalculated. The goal would be to at least have a visible shadow when possible.

source/client/ui/story/CreateLightMenu.ts
```diff
@@ -0,0 +1,115 @@
...
+            this.requestUpdate();
+        } else {
+            this.dispatchEvent(new CustomEvent("confirm"));
+            this.remove();
+        }
+    }
+
+    protected firstConnected() {
+        super.firstConnected();
+        this.classList.add("sv-option-menu", "sv-light-menu");
```
gjcope, line 52: Is sv-light-menu mean to be defined in this PR? I'm not seeing it and the styling of this menu is odd.

gjcope:
@carschno Thanks for the PR and apologies for the delay in review. Looks great!

See the inline comments for a couple things that jumped out and also take a look at improving the keyboard accessibility of the popup menu. Typically we automatically transfer focus to the popup window and then capture it. You can see [focusHelpers](https://github.com/Smithsonian/dpo-voyager/blob/master/source/client/utils/focusHelpers.ts) for a helper function.

Other than that I think it's just a matter of better integration with the UI. Currently it seems a bit awkward. With the current implementation I would expect, as you've noted, that a lighting task would provide access to all lighting setup options. Additionally, we will need to integrate this with the environment lighting updates coming (ideally) in the next release ([dev-shading](https://github.com/Smithsonian/dpo-voyager/tree/dev-shading)) which may actually be more intuitive than the current implementation (a new env light object) but would definitely require a more comprehensive light task.

I'm also a little concerned about the volume of tasks we have now in the UI. Our upcoming UX overhaul is limited to Explorer, so we're attempting to manage complexity as best as possible here.

source/client/components/CVLightsTask.ts
carschno: > Shadow properties should also be copied (where appropriate) and/or recalculated. The goal would be to at least have a visible shadow when possible.

As far as I can see, shadow properties (enabled, blur, resolution) are copied automatically to the new light. Not sure how to deal with light-specific properties. For instance, directional light has a shadow property `size`. When I convert it to a point light and back, the size is lost.

Could you point me to properties that are lost (and/or how to deal with light-specific shadow properties)?

carschno: > I also wonder about the copying of intensity values as we are working with physical units in some, but not all, cases which can make it seem as if a light isn't working when switching...

In the `Light`-related classes, `CDirectionalLight.ts` has been the only one in which I could see any specific handling of the intensity:
```typescript
light.intensity = ins.intensity.value * Math.PI;
```
That is why I added this particular line.

source/client/ui/story/CreateLightMenu.ts
carschno: This was supposed to be `sv-create-light-menu`, which is declared in this file. Fixing, but the style seems to remain unchanged :-/

carschno:
Just to summarize the remaining tasks:
- [ ] Keyboard accessibility popup menu to create lights, transfer focus to popup
- [ ] UI: integrate new options for light creation, deletion and type change with other light settings

gjcope:
@carschno The current commit has build issues for me. It looks like the style migration update from https://github.com/Smithsonian/dpo-voyager/commit/d0901d905ea7983acd473ccb7bc9969d220bf012 is missing for your added color declarations. It should be something like common.$color-text in place of $color-text.

carschno:
I merged in the upstream `master` branch and adapted the styles accordingly, thanks for the pointer!

What I am still missing apparently is a way to populate the values of new lights with defaults. @gjcope would you have another pointer on where to find this?

Once this is done, I think this PR is ready for review.

source/client/ui/story/SettingsTaskView.ts
```diff
@@ -155,4 +215,37 @@ export class SettingsTree extends Tree<ITreeNode>
...
+    // Convert intensity to/from legacy light types
+    // Source: https://sbcode.net/threejs/lights/
+    const legacyLightTypes = [CAmbientLight, CDirectionalLight, CHemisphereLight, CSpotLight];
+
+    const sourceLightIsLegacy: boolean = legacyLightTypes.some(lt => sourceLight instanceof lt);
+    const targetLightIsLegacy: boolean = legacyLightTypes.some(lt => targetLight instanceof lt);
+    if (sourceLightIsLegacy && !targetLightIsLegacy) {
+        targetLight.ins.setValues({ "intensity": sourceLight.ins.intensity.value / Math.PI });
+    } else if (targetLightIsLegacy && !sourceLightIsLegacy) {
+        targetLight.ins.setValues({ "intensity": sourceLight.ins.intensity.value * Math.PI });
```
carschno, line 242: Two questions for review:
1. does it make sense to divide/multiply intensity by PI when converting from/to legacy light types?
2. Is the list of legacy light types correct?

gjcope:
Made a few small tweaks at 0f1ad82 to better align icons and generalized 'canDelete' flag.

> I'm not sure what you mean here exactly. Keep the "add" and "remove" functionalities, but not the "change type" functionality?

Yes, that's what I was considering.

> I assumed that the option for changing the light type would be a useful feature for several users because it was mentioned multiple times (also in https://github.com/Smithsonian/dpo-voyager/issues/190). Adding it on top of the add/remove functionality does not seem to cause too many issues. Or have you noticed issues other than the ones you mentioned above?

I could be wrong, but my impression was that what people really wanted was the ability to configure their own lighting, especially using light types other than directional. I'm not sure they knew add/delete was an option...

I think "change type" works, but I see some clunkiness with it on both the front and back end. One the front, we have the issues with intensity/shadow transference. Some lights have physical units, some have falloff, etc., which I think can make the differences in effect a little confusing.

On the backend we have scenarios like the [SettingsTask](https://github.com/Smithsonian/dpo-voyager/pull/374/files#diff-4f65ddd656cc2472036781f3dfee81622254e4d4c172683b479810a8d42f350f) update where light specific code needs to be inserted into a generalized class to get it to work. Ideally, the settings task would only be showing the defined object properties.

It just seems like maybe too complicated for the convenience factor, but I'm definitely open to arguments. Maybe @sdumetz will weigh in if he has a chance.

gjcope:
A few more tweaks in 8b200e6035280787ff50c3243c99933882518d9e

- moved 'name' property to the light objects
- fixed issue where new light helpers (particularly directional helpers) don't initialize properly when created via UI

I polled folks on our end about changing light type and the consensus was that it seemed like a kind of confusing end user experience at this point. I think there are a number of specifics to work out with how things like light intensities, falloffs, and shadows params get transferred. I'm happy to work through those with you now (Blender does it pretty seamlessly, so I know it's possible!), but also think we could push out the add/delete (which would a big step) and then add in the change-type convenience when time allows.

carschno:
The code looks much cleaner indeed now. Does this resolve your concerns?

PR #377, opened 2025-09-04
Async validate

sdumetz:
As discussed in https://github.com/Smithsonian/dpo-voyager/discussions/317, I want to speed up initial scene load by moving the scene validation off the main thread.

This change moves document validation over to a Web Worker. The gain is twofold: 

#### schema-parsing (which is quite CPU-heavy) should now happen off the main thread

main-thread parsing took ~200ms for a simple scene

starting the worker seems to take no measurable time. further work is non-blocking

#### schema data and the validation library are no longer included in the main bundle

Loading is asynchronous and the main bundle is 140kb slimmer.

### Optimization

If we want to fully replicate existing behavior it is possible to force the application to crash when schema validation fails by throwing an error on failure. Not sure a crash is desirable as long as we report the error.

For the error reporting I'm thinking a Notification might not be visible enough and it may be preferable to open a dialog box? The error messages can get quite long and it might be preferable to have some space to show them without having users open the browser console?

Here we are starting a Web Worker once and it will stay idle forever. We could instead start a short-lived worker on each `openDocument` call. Not sure what would be best overall.

### DIFF
--- a/source/client/components/CVDocument.ts
+++ b/source/client/components/CVDocument.ts
@@ -34,12 +34,14 @@ import CVAssetManager from "./CVAssetManager";
 import CVAnalytics from "client/components/CVAnalytics";
 import { ELanguageType } from "client/schema/common";
 import CVModel2 from "./CVModel2";
+import Notification from "@ff/ui/Notification";
 
 
 ////////////////////////////////////////////////////////////////////////////////
 
 export { IDocument, INodeComponents };
 
+let _worker:Worker;
 
 /**
  * A Voyager document is a special kind of graph. Its inner graph has a standard structure, and it can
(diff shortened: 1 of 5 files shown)


gjcope:
> As for supporting a one-file import, this implementation still only requires only one <script>, even though it loads two files, but we already dynamically import the draco and basis assemblies, the language/*.json files, the logos,... so I expected everyone to already be importing the whole dist/ folder.

It's not about dynamically importing files. It's about using a relative path that assumes they are loading additional files from the same root as the main JS. This is why we have config options like [dracoRoot](https://smithsonian.github.io/dpo-voyager/explorer/api/) and [resourceRoot](https://smithsonian.github.io/dpo-voyager/explorer/api/) which have been specifically requested in the past. Some users prefer to include a link to the 'latest' js from our API, but may have their own resource directory with custom logos, etc. In this validation case it's just about not breaking this flexibility for the existing use cases.

sdumetz:
Ok that should work but to be honest I haven’t tested.

I’ll test this, and check if I can improve the error reporting.

The errors are normally short except if you somehow manage to have a really broken document. In this case the list of errors can get arbitrarily long and we have to handle that in a ok-ish way.

sdumetz:
I fixed the messages that sometimes came out of order. 

[63b6ba0](https://github.com/Smithsonian/dpo-voyager/pull/377/commits/63b6ba0e05e3e4c0d736b38d058e643410078462) uses an even more efficient version that delegates the module loading to the Worker with a workaround for the Same-Origin policy (unless an explicit CSP that prevents it is set on the server).

Let me know if it would still cause problems for some people.

gjcope:
I was able to take a closer look and I think there is still a miscommunication. Functionally everything still looks great. 

My issue isn't specifically to ensure that you can still load that assets folder from a different location than the js, it's ensuring that the main js bundle doesn't *require* any additional files to be co-located to run. With the new validate module, we are creating a dependent requirement.

Ideally, creating a new 'independent' module would be handled like the draco/basis modules. Static files that are loaded from a universally accessible default location, or via a specified alternate root. But in this case that wouldn't really work well as the validation is itself dependent on schemas that change pretty regularly.

I don't have a good solution at the moment but I will continue to explore it.

sdumetz:
We decided to move on with this on our own. I understand this isn't aligned with your vision of how voyager should work at the moment and the changes are purely functional so I don't think it's a huge issue that we diverge on that.

gjcope:
I'd love to find a workable solution to this, but understand the reasoning. Thanks for following up.

PR #378, opened 2025-09-10
Granular snapshots properties selection

sdumetz:
I've got two issues with the current system that selects snapshot properties by "feature-group":

- snapshots can get unreasonably large in large scenes. For example if I have a scene with 100 models, I either can't move models in tours or add a lot of content to my `scene.svx.json`. easily a few megabytes. In those cases I generally only want to move a few of those models, so I'd like a more granular selection to be possible.
- users not familiar with the software often have trouble guessing what those mean. Having a list of included properties for each category might help them.

Here is a screenshot of my initial proposal:
image

It doesn't work for real at the moment: I'll wire things up when the UI is done.

Ideally, I would also want to address #255 and possibly #379.

### DIFF
--- a/source/client/ui/story/ToursTaskView.ts
+++ b/source/client/ui/story/ToursTaskView.ts
@@ -51,8 +51,7 @@ export default class ToursTaskView extends TaskView<CVToursTask>
 
         return html`<div class="ff-scroll-y">
             <div class="sv-commands">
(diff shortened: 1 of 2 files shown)


gjcope:
> In both those cases, with this proposition you can select only the few properties for the few models you need to animate, instead of 22 props times however many models you have.

I'm still unclear on this as the screenshot only seems to show toggling properties and not objects, but I haven't looked at the code yet so I'm sure that will clear it up. I'll follow up after I'm able to take a look. Thanks!

PR #382, opened 2025-09-12
conditional beforeunload warning, fix #205

### LINKED ISSUE #205: Fire "beforeunload" confirm box only if document has changes
This box is annoying. The save process is complicated enough (save scene, capture pose, save article), if users can't rely on this to _know_ if they forgot to save something they will just ignore the dialog because it's wrong most of the time anyway.

Serializing the `CVDocument` just after `openDocument()` and serializing again to compare the string value in `CVStoryApplication.beforeunload()` might do the trick but I'd like a solution that allows a better UI (eg: highlighting the **save** button as soon as the doc is changed). The tricky bit is that we have no representation of the document's data once it's been loaded into the graph, so a "changed" switch seems hard to implement.

I'd like to work on this issue, if anyone has an idea worth trying.

sdumetz:
Those changes should allow us to only display the "Exit Confirmation" box if the document has been changed, when the user navigates away from Voyager Story.

Functionally, it has two parts:

- some code to store the initially loaded document and deep-compare the current `IDocument` (`CVDocument`'s serialized state) to it.
- some adjustments to prevent state serialization from sending unnecessary changes (eg: Camera Coordinates, as they are always derived from `Navigation.Orbit` and `Navigation.Offset`).

I'm a bit concerned about the code for the `deepEqual` function that seem a bit too _ad hoc_ for my taste.

I've temporarily modified the `Document.Download` callback in `CVStoryAplication` to test things more easily,  that would need to be cleaned up once everything else is OK.

### DIFF
--- a/libs/ff-ui/source/styles/_styles.scss
+++ b/libs/ff-ui/source/styles/_styles.scss
@@ -465,6 +465,11 @@ button, input {
       fill: $color-icon-disabled;
     }
   }
+
+  &.ff-info { color: $color-info; }
+  &.ff-success { color: $color-success; }
+  &.ff-warning { color: $color-warning; }
+  &.ff-error { color: $color-error; }
 }
 
 .ff-menu {
--- a/source/client/components/CVDocument.ts
+++ b/source/client/components/CVDocument.ts
@@ -43,6 +43,38 @@ import CVModel2 from "./CVModel2";
 export { IDocument, INodeComponents };
 
 
+function deepEqual(path:string, x: any, y: any):boolean {
+    if (x === y) {
+        return true;
+    }else if(typeof x === "number" && typeof y === "number"){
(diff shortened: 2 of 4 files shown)


gjcope:
> Not sure how this could be elegantly handled. Bumping the scene file's version number and putting this code behind a min-supported-version condition?

> If this still happens after you have saved over the scene though, it's a bug.

No, I think it's exactly what you describe but I will try out a slightly different approach when I have a few cycles.

> Me neither. I wanted to prototype it so we can test how it would work. And for debugging it's nice to have it. But I'm not sure it should stay, especially if have no workaround for the "first save is always dirty" issue. If we remove the highlight, having all old scenes considered dirty would simply match the existing behavior, so it shouldn't be too confusing.

Agree 100%

PR #401, opened 2025-12-08
Extract PropertyPercent

carschno:
## Motivation

As noticed in https://github.com/Smithsonian/dpo-voyager/pull/391#issuecomment-3564253974, numeric input fields use `type="text"` to facilitate displaying percentages. However, this prevents the use of standard browser input methods for regular numbers, which poses problems for accessibility and on mobiles.

#391 adds an input field for date and time which makes the use of `type="text"` particularly inaccessible.

Therefore, this PR extracts the logic for distinguishing between regular numbers and percentages from `PropertyNumber` to `PropertyPercent.ts`.

## Changes

- Create new `PropertyPercentage.ts` type, implementing the custom element `sv-property-percent`
- Remove the conditionals for `schema.percent` from `PropertyNumber`, and use `type="number"` for input fields.
- Add a style for `sv-property-percent` (extending `sv-property-number`)
- Define the step size for numeric component properties to be used by the browser input methods

image

Numeric input fields now correctly use `type="number"`, automatically enabling the appropriate input methods in the browser. The percentage input fields are unchanged.

### DIFF
--- a/libs/ff-scene/source/components/CCamera.ts
+++ b/libs/ff-scene/source/components/CCamera.ts
@@ -34,11 +34,11 @@ export default class CCamera extends CObject3D
         rotation: types.Vector3("Transform.Rotation"),
         order: types.Enum("Transform.Order", ERotationOrder, ERotationOrder.ZYX),
         projection: types.Enum("Projection.Type", EProjection, EProjection.Perspective),
-        fov: types.Number("Projection.FovY", 52),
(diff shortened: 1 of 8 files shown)


carschno:
I noticed a flaw on enum types: they are now shown as a number input type, whereas they should be a dropdown.
image

Same view on current `master` branch (correct):
image

carschno:
Closing, not doing.

PR #409, opened 2025-12-31
Tag and batch-enable lights

carschno:
# Changes

- Add tags property to lights
- Add selection checkbox for light nodes for batch enabling/disabling
- Add tag-based light node filter to select multiple lights based on tags

# Screenshot

image

Closes https://github.com/Dynamic3D-Project/dpo-voyager/issues/22


gjcope:
Spoke with @kgschoueri and I think this is a misunderstanding. The tagging needed is the same tagging that currently exists for models and annotations, just extended to lights.

PR #423, opened 2026-01-26
Translate the "+more info" annotation string

Inao0:
Very similar to #420 except the translation does not exist in the other languages yet. I translated it in French.
The "+ more info" string appears when "pushing" an extended annotation against the side of the screen.
However I could not get the circle annotation to display this string :
The "+more info" is in its code,  but when pushed against the side of the screen, the annotation flips toward the center instead of shrinking an displaying "+more info".
Maybe both types of annotation should have the same behavior ?


gjcope:
Merged to [rc-58](https://github.com/Smithsonian/dpo-voyager/tree/rc-58)

PR #427, opened 2026-02-09
HDRi Map Upload

### LINKED ISSUE #35: HDRi Environment Maps
See Pull Request for details: https://github.com/Smithsonian/dpo-voyager/pull/427

Link to [mock-up](https://excalidraw.com/?element=VMapxJzPMWwNEH77g1qi6#room=e7d7b1e4c4fe079a89ca,f2iJytAs6Hg0qbSIHk7Tgw)

Closes #23 (duplicate)

carschno:
Closes https://github.com/Dynamic3D-Project/dpo-voyager/issues/24
Overrides https://github.com/Smithsonian/dpo-voyager/pull/411

# UI/UX Changes

- When a user uploads an `.hdr` file (via drag&drop), it is added to the list of selectable environment image maps.
- When a user deletes an `.hdr` file (via the asset/media manager), it is removed from the selectable environment image maps.

# Implementation

- `CVEnvironment.ts`:
  - avoid glitches for uploaded maps: re-render upon update and track request id
  - add `addImage()`, `deleteImage()` and `_updateImageIndex()` for (un-)registering image maps
  - (no functional change) remove spurious whitespaces throughout file
- `CVMediaManager.ts`: (un-)register environment maps when deleted or uploaded.

# Preview

This [preview](https://dpo-voyager-git-feature-hdriupload-carstens-projects-7b21f278.vercel.app/voyager-story-dev.html?document=https%3A%2F%2Fyhp1co7ikmvyqrk7.public.blob.vercel-storage.com%2Ffiles%2Fmodels%2Ftusk%2Ftusk.svx.json) illustrates the changes, but cannot provide a file backend so that the upload functionality cannot be tested

https://github.com/user-attachments/assets/c64a6162-c8bc-4f8e-9a78-14678ef3d3b3


carschno:
From initial user testing:

> Yeah, things kind of worked when I refreshed the link. But each time I do a refresh, there is something slightly off. It was first the map mixing and now it seems to be that the hdr map is showing as a box for the 3,4 and 5 maps. but then now I can see the 0,1 and 2 in a normal way. 
 
> It also seems to have mixed up the maps...what should be 5 is now 4. And 5 is I think the original number 3. 
 
- [x] switching maps causes artifacts in background
- [x] fix map ordering

gjcope:
@carschno apologies for the delay in getting to this.

Can you say a bit more about the reasoning behind the image scan? It seems like the intent might be to have additional (or different) default options. If so, maybe this could be handled more like the logo/spinner solution that requires whoever is deploying story to replace files in the assets folder? I'm not sure we want any user to be able to add defaults.

For scene-specific maps, they should be assignable when added and then saved/loaded via the scene so I'm not sure a scan is necessary. If we add an image-only drop handler to the media manager we could have a pop-up that lets you assign the purpose of the image. That way we could also accommodate adding overlay maps and any future image-only content.

carschno:
@gjcope No worries, and my apologies in return for another delay on my side.

The ability to upload new maps was a feature requested by @kgschoueri (cf. issue https://github.com/Dynamic3D-Project/dpo-voyager/issues/24). I see your concern about users uploading files; do I understand correctly that they should be located at the same server as the scene file instead? I suppose @kgschoueri might have had a workflow in mind, but I cannot tell at the moment.

I also realize that this PR might conflate the tasks noted in the issue linked above. I will think about compartmentalizing them more clearly.

gjcope:
Hey @carschno ,

I totally get the reasoning and requirement for uploading new environment maps, just unsure about the reasons behind the "image scan" piece.

I think we have two situations:
1. Default maps. We definitely need to ship defaults, as many users won't need or care about a custom map. I think the issue is if someone deploying their own instance wants different defaults. That was why I mentioned the logo/spinner scenario as that is something we address there. An "image scan" of the assets directory could be an alternate approach for this, but would be difficult unless we are only supporting a format that clearly maps to env maps (like HDR).
2. Custom maps. We have current use cases (and I'm sure there will be more) where users want a specific custom map for a specific object. This is what I think would be handled by drag/drop. But we also have other use cases for uploading custom images (like overlays) so I think this is an opportunity for a more general solution.

> do I understand correctly that they should be located at the same server as the scene file instead?

The way it currently works, any scene specific (non default) assets that are ingested are added to the scene package. That ensures the archivability of it. In tandem to this I am also working on a IIIF compliant branch that supports dragging/dropping URIs as well since that is a requirement for IIIF manifests.

carschno:
Hi @gjcope ,
I think that I get your points, now I am trying to map them to a UI-based abstraction before I re-try an implementation.

1. For the default maps, I suppose we do not need a change then. I understand, the default image files are the ones listed in `CVEnvironment.ts`. Parametrizing them seems like a potential improvement, but I don't see a need currently.
2. Custom maps are harder. I have a few doubts:
    1. Just to confirm: this affects only Story, right?
    2. How should the upload work exactly? Do you mean that the drag & drop-upload should work here in the same way as in my PR (ie. through the asset/media manager interface, or should we try to design a new approach? Perhaps adding a dialogue that lets the user specify the use of the file (e.g. environment map or overlay)? This would make the file scan unnecessary indeed.
6. Also UI-related: should we add a separate selection box for the non-default environment maps, e.g. a drop-down for the scene-specific files? I am torn; on the one hand, the distinction between default and custom maps seems irrelevant when selecting a background. On the other hand, mixing the custom maps into the default maps would require a more intuitive UI than the current indices because they will be very hard to understand with a dynamic number of files.

If you have good ideas and/or specific requirements, I am looking forward to any input!

Update: I might not fully understand what you mean by scene packages? Is there a specification you could point me to?

gjcope:
Hi @carschno ,

1. Sounds good. I think wanting to modify the defaults is a totally valid use case, but if it's not a concern for you at the moment I'm fine with leaving that out.
2. Custom maps
    i. Yes. Explorer already has the support for reading custom maps from the scene file.
    ii. Yes, exactly. I think we would want UI, similar to what you get when drag/dropping a model, to let you specify the usage of the image. The only inconsistency I see is with assets like audio and annotation images where you drag/drop directly to a text box that is in-context. In these situations the usage is inferred. But with the environment maps we have the default/custom separation and also wouldn't be able to support more than one custom map very well (that I can think of).
3. I don't think we need two separate UI elements, but you've highlighted another important issue. In my opinion, even the defaults are difficult to understand as indices, but I don't have a fix at the moment. Ideas definitely welcome. We use full filenames for overlay maps, but that can get unwieldly. It also translates to Explorer as it's difficult to fit longer names in that UI. But personally I feel like the clearness of the indices is less important on that end.

> Update: I might not fully understand what you mean by scene packages? Is there a specification you could point me to?

A scene package just refers to the collection of assets needed to display a Voyager scene. So this would include the scene.svx file, the 3d model assets, and any content assets (articles, images, audio, etc.). This is what you get when you click "download" in Story Standalone mode, a zip with all of these assets. For custom environment maps, the map would become another one of these assets (similar to overlays).

carschno:
@gjcope @kgschoueri I have updated the PR and the description. Please let me know if this is in line with your thoughts!

gjcope:
Hey @carschno , thanks for the updates!

- Currently the "MapPath" property does not show up for me in the Explorer UI (just the label shows), but I'm not sure how well it would work there if it did. This was why we initially went with the un-ideal index approach. Map paths are generally pretty long and also not really relevant to the end user. Maybe we could truncate it to just the file name? Or provide a configurable alias that could be character limited? Not sure what the best solution is yet but it's important that the front end UI remains clean and usable.
- The drag/drop works to add a new map to the package, but it doesn't assign it as an environment map so is not available via the index selection. On drag/drop of an image, if we had a pop-up that allowed you to specify that the image is meant for (env map, overlay map, etc.) this could be a pathway for setting it up for use.
- Changing what gets serialized/loaded means it will not be backwards compatible which we would like to avoid if possible.
- I have a couple questions about the loading code changes but will follow up here after I think through the logic a bit more.

carschno:
Thanks for the feedback @gjcope !

> Currently the "MapPath" property does not show up for me in the Explorer UI (just the label shows), but I'm not sure how well it would work there if it did. 

What do you mean it does not show up? Is it related to screen size or is there a bug I have not noticed?

Regarding the index dropdown (options) selector: I think the current solution is manageable with a fixed number of images, but becomes unworkable with custom maps.

Truncating `mapPath` to make it more readable seems reasonable, although potentially ambiguous. It seems rare, but I could imagine a user uploading an updated version of a default map with the same name, for instance. Perhaps the risk is tolerable, though.

Anyway, my assumption has been that users normally populate the `mapPath` text field via drag&drop (or via the index dropdown for the default images).
This implies that the text field could be replaced by a dropdown that lists the available maps, combining default and custom maps. However, I tried that before by scanning the file system to find all custom maps, but that was not ideal due to the dependencies on WebDAV, predefined filesystem structure, and file name extensions.

When you mention assigning it as an environment map, however, that sounds promising. Are such assignments done for other file types? Do you have a pointer?
It still poses the issue of finding previously uploaded custom maps, but that seems solvable.

> Changing what gets serialized/loaded means it will not be backwards compatible which we would like to avoid if possible.

Noted! Probably best to keep this in mind while finding a good solution for the above. What is your stance on keeping the `imageIndex` as an optional field when reading scene files, but deprecate it while introducing a new field (`mapPath` or similar)?

gjcope:
> 
> What do you mean it does not show up? Is it related to screen size or is there a bug I have not noticed?
> 
Only the label shows in the UI:
image

> Truncating `mapPath` to make it more readable seems reasonable, although potentially ambiguous. It seems rare, but I could imagine a user uploading an updated version of a default map with the same name, for instance. Perhaps the risk is tolerable, though.
> 

At some level we can't stop people doing weird things if they choose to, but I think as long as standard/common use cases work well then we are fine.

> Anyway, my assumption has been that users normally populate the `mapPath` text field via drag&drop (or via the index dropdown for the default images). This implies that the text field could be replaced by a dropdown that lists the available maps, combining default and custom maps. However, I tried that before by scanning the file system to find all custom maps, but that was not ideal due to the dependencies on WebDAV, predefined filesystem structure, and file name extensions.
> 

My main issue there is just that dragging and dropping a new file does not make it available as a map, it just adds it to the scene package.

> When you mention assigning it as an environment map, however, that sounds promising. Are such assignments done for other file types? Do you have a pointer? It still poses the issue of finding previously uploaded custom maps, but that seems solvable.
> 

Sure. Take a look at the onMeta handler here: https://github.com/Smithsonian/dpo-voyager/blob/57c96d6f1c79c73b7aca65795c0ab5841dc21604/source/client/components/CVEnvironment.ts#L253
We are looking for any custom maps defined in the scene file as environment maps and then assigning them to the option list. This is also how previously uploaded custom maps would be handled. They get saved to the Meta component as env maps.

> > Changing what gets serialized/loaded means it will not be backwards compatible which we would like to avoid if possible.
> 
> Noted! Probably best to keep this in mind while finding a good solution for the above. What is your stance on keeping the `imageIndex` as an optional field when reading scene files, but deprecate it while introducing a new field (`mapPath` or similar)?

I'm fine with it as long as we have a reasonable solution for the front end UI to display the names. So far I haven't been able to think of one.

carschno:
Regarding the invisible string input field, I had forgotten to create the CSS. Fixed.

> We are looking for any custom maps defined in the scene file as environment maps and then assigning them to the option list. This is also how previously uploaded custom maps would be handled. They get saved to the Meta component as env maps.

This was not entirely clear to me, thanks! Just to confirm, I understand the current logic is that
environment maps are defined in `CVEnvironment._imageOptions`as combination of those in `CVEnvironment.images` (default) plus all maps defined in the `meta` sections in the scene file (custom). The index drop-down is a reference to the `_imageOptions`. Correct?

In order to extend the list of available images with uploads, we should:
1. enable upload for `.hdr` files (done in this branch already)
2. when an `.hdr` image is uploaded:
    3. add it to `Meta.images`
    4. add it to `CVEnvironment._imageOptions` (like in `onMetaComponent()`)

Consequently, uploaded images will be serialized in `CVMeta.toDocument()`. Through `fromDocument()`, those uploaded images will be loaded according to the current logic when the scene package is loaded.

Regarding this, I wonder how the current index-based solution has been motivated. Why not use a drop-down for the image names? As discussed above, file name lengths (with paths) might be a UI issue, but that could be mitigated by truncating as we discussed.

Perhaps this would be an occasion to implement this change? Instead of introducing the string input field (like I have done in this PR), replace the existing index drop-down with a filename drop-down?

gjcope:
> This was not entirely clear to me, thanks! Just to confirm, I understand the current logic is that environment maps are defined in `CVEnvironment._imageOptions`as combination of those in `CVEnvironment.images` (default) plus all maps defined in the `meta` sections in the scene file (custom). The index drop-down is a reference to the `_imageOptions`. Correct?

Yes, correct.

> 
> In order to extend the list of available images with uploads, we should:
> 
> 1. enable upload for `.hdr` files (done in this branch already)
> 2. when an `.hdr` image is uploaded:
>    3. add it to `Meta.images`
>    4. add it to `CVEnvironment._imageOptions` (like in `onMetaComponent()`)
> 
> Consequently, uploaded images will be serialized in `CVMeta.toDocument()`. Through `fromDocument()`, those uploaded images will be loaded according to the current logic when the scene package is loaded.

As long as the correct ImageUsage type is also set ("Environment") then yes, that sounds correct.

> 
> Regarding this, I wonder how the current index-based solution has been motivated. Why not use a drop-down for the image names? As discussed above, file name lengths (with paths) might be a UI issue, but that could be mitigated by truncating as we discussed.
> 
> Perhaps this would be an occasion to implement this change? Instead of introducing the string input field (like I have done in this PR), replace the existing index drop-down with a filename drop-down?

I think the original motivation was mostly UI based as well as thinking about future-proofing changes to defaults. If we decided to replace a default map with an improved one, it would need have exactly the same name. 

I'm definitely fine with a name drop-down taking the place if the index list in Story config, I think the issue is with the Explorer UI and what gets serialized. We aren't currently using drop-downs in the frontend UI as the buttons are a better user experience on mobile. But we can definitely test it out with users and see how it is received. I'll need to consideration the serialization approach a bit more as I think there are pros and cons to each (index or filename).

gjcope:
Also noting, not a requirement for this PR, but I think an ideal solution for uploading would be a pop-up that lets you pick the usage of the image. It's pretty obvious with .hdr files, but this mechanism could also be used to upload and assign overlay images or even to create image planes like several Pure3D folks have manually created.

carschno:
I have redone this from scratch after getting a better understanding of the inner workings. Skipping the UI changes for now, uploaded hdr maps are now registered and added to the list of selectable environment maps.
I have skipped the pop-up for the purpose for now, as I think it should be considered in the broader context of uploaded files, whereas it is indeed obvious for hdr images.

An aspect still be to considered is the handling of .hdr files that exist in the filesystem, but are not defined in the scene file. However, I am not sure if this is an expected situation, hence whether it needs to be handled.

carschno:
The serialisation functionality is related to https://github.com/Smithsonian/dpo-voyager/issues/493

gjcope:
Hey @carschno - thanks for the update! I've made some additions/adjustments here: https://github.com/Smithsonian/dpo-voyager/tree/dev-hdri-upload Let me know if anything seems to be off-base.

> I have redone this from scratch after getting a better understanding of the inner workings. Skipping the UI changes for now, uploaded hdr maps are now registered and added to the list of selectable environment maps.
> I have skipped the pop-up for the purpose for now, as I think it should be considered in the broader context of uploaded files, whereas it is indeed obvious for hdr images.

I think it's fine to skip the pop-up for now, but noting that even with environment maps there is the possibility of extensions like jpg (https://developer.android.com/media/platform/hdr-image-format) that would need disambiguation so I think it is a must at some point. It will also support other features that I know @kgschoueri and co would appreciate like being able to drag/drop an image and have it generate an image plane in 3D space.

A few comments/questions:

> CVEnvironment.ts

- Can you say a little more about the issues you were seeing that necessitated the request id addition? I'm not seeing the logic  improvement but am probably missing a use case.
- In the branch above I commented out a couple of things that seemed redundant. Let me know if I am missing something important.
- In order for the asset to get saved to file, an image asset needs to be created and added to the meta array. I made a first pass at this, but it is still missing accurate image info. It supports the proof of concept for saving/loading though.

> CVMediaManager.ts

- It looks like this was implemented only for the upload path via the media tab. We have two other paths (standalone and enabling the drag/drop attribute) that need to be handled as well. This should be working for adding files in the above branch, but I have not yet added deleting.

carschno:
> Can you say a little more about the issues you were seeing that necessitated the request id addition? I'm not seeing the logic improvement but am probably missing a use case.

Yes, although it is hard to reproduce. @kgschoueri noticed a while ago that enabling custom HDRi images did not always work properly, leaving visible artifacts in the map. I have seen this a few times myself, but could not reliably reproduce this, unfortunately. My guess is that it is related to the loading (network) speed and concurrent updates of the environment map. Therefore, I added the request id to make sure that concurrent changes do not interfere with each other. As far as I can tell, the issue has since been resolved.

BTW, I have added an attempt to read the HDR file header in [this branch](https://github.com/Dynamic3D-Project/dpo-voyager/tree/feature/hdri_upload_read_hdr) in commit 4eee277dc93be4e81d79e395689086635230c8be. It currently does not seem to work because the filepath is resolved to e.g. http://localhost:8000/articles/Media/scythian_tombs_2_4k.hdr for `articles/Media/scythian_tombs_2_4k.hdr`, but this is not correct. Anyway, the method might or might not have to be fixed according the the other upload paths you mention above.

source/client/components/CVEnvironment.ts
```diff
@@ -259,14 +285,53 @@ export default class CVEnvironment extends Component
...
+            const meta = this.getGraphComponent(CVMeta);
+            if(meta) { 
+                meta.images.insert({
+                    uri: image_uri,
+                    usage: "Environment" as TImageUsage,
+                    quality: "HDR" as TImageQuality,
+                    byteSize: 4096,
+                    width: 1024,
+                    height: 1024,
+                }, "HDR");
```
carschno, line 314: @gjcope I don't understand why you set the id to `HDR` here. As a consequence, the next uploaded HDR image will overwrite this one in `meta.images`. Is this intentional?
Intuitively, I would have used the uri as a id as a unique, distinguishing feature.

carschno: I understand that this is related to (de-)serialisation. I understand that:
- the id is not serialised.
- upon deserialisation, the image quality (`HDR` in this case) is used as id (also to be used for deletion)

Anyway, it still looks like we cannot serialise multiple environment images.

gjcope: Hi @carschno , that's all correct. Right now there is a single slot for a custom environment map that gets combined with the defaults. This makes the frontend UI situation predictable and manageable. We would need to refactor the UI (as we have previously discussed re: moving away from indices) for it to be feasible to support unlimited custom maps.

carschno: Ok, thanks! I am pondering the consequences in this scenario:
1. user uploads custom image `img1.hdr` -> is added to `metas[].images` with id `HDR`
1. user uploads custom image `img2.hdr` -> is added to `metas[].images` with id `HDR`, hence removing `img1.hdr`
2. user removes `img2.hdr` -> is removed from `metas[].images`

Should there be a fallback for `img1.hdr`? Anyway, I don't see a way of implementing this without the risk of unexpected behaviour, so I would keep it as it is, unless I am missing something. I wanted to raise this question anyway.

carschno:
> It looks like this was implemented only for the upload path via the media tab. We have two other paths (standalone and enabling the drag/drop attribute) that need to be handled as well. This should be working for adding files in the above branch, but I have not yet added deleting.

I tested this a bit and I understand that the two other paths are now both covered through `CVMediaManager.ingestFiles()` (79444cd948dacb547d7c4493e1c7061b96a7ee01).

I have also added `CVEnvironment.readHdrDimensions() to get the HDR image dimensionality.

@gjcope Do these additions sufficiently address your earlier concerns?

gjcope:
@carschno Thanks for the updates and clarifications. I think we are getting close. I would still like to take a look at the loading issue above to see exactly what is at play. It's not clear to me what the request id is fixing, but I will give it another review.

I've also updated https://github.com/Smithsonian/dpo-voyager/tree/dev-hdri-upload. It's still very much WIP but has the selection menu functionality I was thinking of. I also updated the image storage so you can add more than one HDRI, but I do think it's a balance between flexibility and a good user experience.

image

gjcope:
@carschno I'm looking at pushing this functionality out next week (with the other image import functions as well). Were you planning to incorporate the move to image names instead of indices or are you all good with it for now? Thanks!

carschno:
> @carschno I'm looking at pushing this functionality out next week (with the other image import functions as well). Were you planning to incorporate the move to image names instead of indices or are you all good with it for now? Thanks!

@gjcope Apologies for the delayed response. For now, I think I would not add changes in the UX to this PR and rather make these changes interatively. As far as I am concerned, your additions in the `dev-hdri-upload` branch are fine, but perhaps also better to be handled in a separate PR.

Regarding the request id: due to the difficulties of reproducing the issue that we observed (artifacts when changing the environment map), I cannot guarantee that this logic really fixes the root cause. Anyway, I do think that tracking the current request should make the loading process as a whole more robust against race conditions due to environment image settings triggered faster than they can be executed.
Please let me know if you have more questions!

gjcope:
> @gjcope Apologies for the delayed response. For now, I think I would not add changes in the UX to this PR and rather make these changes interatively. 

Sounds like a good plan.

> 
> Regarding the request id: due to the difficulties of reproducing the issue that we observed (artifacts when changing the environment map), I cannot guarantee that this logic really fixes the root cause. Anyway, I do think that tracking the current request should make the loading process as a whole more robust against race conditions due to environment image settings triggered faster than they can be executed. Please let me know if you have more questions!

I think what I'm missing is how the request id check differs from the mapIdx check that was already there.
`if(requestId === this._loadRequestId && mapIdx == ins.imageIndex.value)`

In both cases it looks like aim is to ensure that the update only happens if the texture being passed in is also the current texture requested by the user. '_loadRequestId' and 'ins.imageIndex.value' both track the last image requested to load, and 'requestId' and 'mapIdx' both identify the texture requested for update. Is there an outcome here that is only handled by request id? I'm happy to go that route, but either way it seems redundant so we should get rid of the mapIdx tracking and check.

carschno:
> I think what I'm missing is how the request id check differs from the mapIdx check that was already there.

The way I understand is that the local `mapIdx` checks whether the map index setting has been changed at all. The global `requestId`, however, keeps track of the loading state. It makes sure that the most recent update to the setting is only triggered if the previous map has finished loading, while the map index setting might already have changed again.

gjcope:
> The way I understand is that the local `mapIdx` checks whether the map index setting has been changed at all. The global `requestId`, however, keeps track of the loading state. It makes sure that the most recent update to the setting is only triggered if the previous map has finished loading, while the map index setting might already have changed again.

I'm not following. `mapIdx` isn't checking whether the map index has changed, it's making sure that the currently requested update only happens if the loaded map is the same as the one indicated by the current map index. 

The `requestId` looks to make sure that the currently requested update only happens if the loaded map is the one that was last requested. 

The only scenario I can think of where this would differ is the race condition where a user has selected a new map such that `ins.imageIndex` changes immediately followed by `updateEnvironmentMap()` getting executed on a previous load before the new `loadEnvironmentMap()` gets called. But in this case `requestId === this._loadRequestId` evaluates to true when it should not.

Do you have a race condition (or other) scenario that the requestId check catches that the mapIdx check does not?

Sorry to drag this out, I just want to understand the issue we're addressing with this code so future updates or optimizations are easier to handle.

gjcope:
@carschno FYI I merged your updates and made a few tweaks at https://github.com/Smithsonian/dpo-voyager/tree/dev-hdri-upload

I know it's a lot, but I need to start merging things for the release candidate.

I did notice that the image dimension function was erroring on fetching "HEAD" when in standalone mode since the url is a local blob. I added a check to skip it for now, but if you know of a solution please let me know.

I also think deleting could be improved/generalized but that could always come later. It seems to be functional for now (at least for the HDRs)

PR #440, opened 2026-03-03
Make sunlight time timezone-aware

### LINKED ISSUE #438: Wrong sun position wrt time, latitude, longitude
# Description

The sun position seems to be set off, which might be related to timezone wrt the latitude/longitude.

> For instance, Washington DC you will see it show light starting at 11am for this time of year which is not correct and light stopping well after it should.

carschno:
Fixes https://github.com/Smithsonian/dpo-voyager/issues/438

# Changes

- add [day.js](https://day.js.org/) dependency
- use Dayjs date objects for timezone-awareness
- add timezone property to `CSunLight.ts` and `CVSunLight.ts`
- add `timezone` to `document.ts` for serialization
- migrate `PropertyDateTime` to Dayjs date objects

image

Note: commit 7aefa4258aa96f8ab5ab68fd26543aad9d244077 adds a timezone field to the light tool, slightly in conflict with related discussions we had earlier. 

However, having the timezone information not visible at all in ~~Story~~ Explorer seems potentially very confusing for users, especially if the coordinates do not align with the selected timezone.
Also, the timezone is now represented in a text field which should cause less difficulties on small screens than the previous dropdown field. 

image


gjcope:
> My point is that the timezone should at least be visible to the user in Explorer to avoid the same confusion that you experienced earlier when the coordinates and the timezone do not match (see original description of #438). This may plausibly happen in scenarios in which the timezone is set to the author's location, but an object is located elsewhere (via the sunlight coordinates). This might be intentional or by mistake, but should be transparent to the Explorer user. The timezone field could be disabled so that Explorer users cannot change it.
> 

I'm fine with there being a label, but I think having a whole input widget is unnecessary, especially if it is disabled anyways. But also, I think the expectation should be that the timezone matches the location. If it doesn't, I would say that is a misconfigured scene.

> Alternatively, option 2 in [#438 (comment)](https://github.com/Smithsonian/dpo-voyager/issues/438#issuecomment-3983693920) (guess the timezone based on the coordinates) could replace manual timezone settings even in Story, but with the chance of inaccurate timezone guesses, this seems awkward.

What if we just went with UTC zones? I feel like that would simplify a lot of things and can be pretty easily calculated from longitude. Combined with the the label, seems like it would work?

PR #445, opened 2026-03-12
Add programmatic annotation focus and stop auto-rotation on focus

ndejardin425:
## Summary
  - Consolidate annotation activation behavior so that **all activation paths** (mouse click, keyboard, API) trigger the same side-effects
  - `setActiveAnnotation(id)` now also animates the camera to the annotation's associated view and stops auto-rotation, making `focusAnnotation` redundant
  - Update API documentation accordingly

  ## Changes

  **Consolidate activation side-effects into `CVAnnotationView`**
  - `CVAnnotationView.ts`: Move camera tween, auto-rotation stop, and orbit normalization from `onSpriteClick()` into the `set activeAnnotation` setter — ensuring consistent behavior regardless of how an annotation is activated
  - `CVViewer.ts`: simplify `activeAnnotation.changed` handler to a simple dispatch

  **Documentation**
  - `api/_index.md`: update `setActiveAnnotation` description to mention camera animation
  - `api-examples/annotations/_index.md`: update description

  ## Test plan
  - [ ] Load a scene with annotations that have associated camera views
  - [ ] Click an annotation sprite — verify camera animates to its view and auto-rotation stops
  - [ ] Call `voyagerElement.setActiveAnnotation("annotation_id")` via API — verify same behavior as mouse click
  - [ ] Verify that activating an annotation without an associated view still works correctly
  - [ ] Verify that clearing the active annotation (empty id) does not stop auto-rotation


gjcope:
> * To be honest, I don't know the project enough. For what I understand, using isInUse would be better, but how can we know when the tween ends ?

I haven't tested it, but isInUse should have the same result, just a little more readable as an indicator that some action has been taken that we are using as a proxy for the user interacting with the app such that the prompt can end. It should ultimately just set the same promptActive property here: https://github.com/Smithsonian/dpo-voyager/blob/a27d2b19d6a24bb776e1076d4eddedbd46f959f6/source/client/ui/explorer/ContentView.ts#L153

I'm probably misunderstanding, but not sure what you mean by "how can we know when the tween ends?" If you can say more, that would be helpful. Which tween and why do we need to know that it ended?

> * You mean if there is a annotation view linked to the annotation, put add an ins that would serve as parameter to know if voyager should trigger the annotation view or not ?

My thought was that someone using the API function may want a choice of whether or not activating an annotation will stop the auto-rotation. So it would likely be a parameter in the API function. But thinking about it some more, that would lead to inconsistent behavior with annotations activated via the UI. I think you can just ignore this. If it comes up as a future use case, I think we could add a global scene setting to indicate if annotation views should interrupt the rotation or not. But we will wait to see if it comes up.

PR #446, opened 2026-03-12
Change tags fields to a specific input

ndejardin425:
## Summary
         
  Replaces the plain text input for annotation tags with a new sv-property-tags custom element that provides a proper tag management UI.
                                                                                                                                                                                                                                                                                                                    
  ## Changes:
  - Adds PropertyTags.ts: a new web component that renders selected tags as removable chips, with a text input (Enter to add) and a dropdown to pick from the existing tag cloud                                                                                                                                    
  - Updates AnnotationsTaskView to use sv-property-tags instead of a plain text field for the tags property                                                                                                                                                                                                         
  - Adds supporting styles in story/styles.scss                                                            
  - Adds i18n strings for the new UI labels (all 7 languages)                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                    
  This improves the story editor UX when managing annotation tags, preventing typos and making the available tags discoverable.


ndejardin425:
> 1. Can you extend this to the other tag fields as well?

I've extended what I've done on ArticlesTaskView, ToursTaskView et SettingsView. Tell me if I missed something ? Do you think it is still required to create property view for the tags ?

> 2. I think this is pretty intuitive when adding new tags, but some people were a bit confused when loading existing tagged scenes as to what the tag widget was. Could we round the corners so that they look more like the tag buttons in the Explorer UI and potentially move the 'x' outside of the box visually? @carschno had a nice style for a tag chip in a previous PR that maybe could be borrowed from.

I've changed the style for the tag field based on the style modifications by @carschno 

> 3. It would be great to change "New tag" to "Add a tag" or "Create a tag" or something more action oriented.

Done

carschno (commented):
This looks very nice indeed. What seems to be missing:
1. `CLight.ins.tags` should use `types.Tags`
2. `CVViewer.ins.activeTags` should use `types.Tags`
3. Consequently, `CVViewer` should also subscribe to `tagCloud`

Additionally, this might be an occasion to use a native `string[]` instead of the comma-separated `string` for the tags (see my in-line) comment.

source/client/ui/story/SettingsTaskView.ts
```diff
@@ -16,7 +16,6 @@
  */
 
 import Node from "@ff/graph/Node";
-import Component from "@ff/graph/Component";
```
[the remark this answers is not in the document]
ndejardin425: yes, the import was not used

source/client/ui/properties/PropertyTags.ts
```diff
@@ -0,0 +1,171 @@
...
+import CVDocumentProvider from "client/components/CVDocumentProvider";
+
+import PropertyBase from "./PropertyBase";
+
+////////////////////////////////////////////////////////////////////////////////
+
+@customElement("sv-property-tags")
+export default class PropertyTags extends PropertyBase
+{
+    type = "string";
```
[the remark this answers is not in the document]
ndejardin425: Might be good idea, but I would prefer to do it in another pull request.

source/client/ui/properties/PropertyTags.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    @property({ attribute: false })
+    protected inputValue = "";
+
+    protected _tagCloudProperty: Property = null;
```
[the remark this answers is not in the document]
ndejardin425: Indeed, you're right, I just kep the name used in the application

assets/language/string.resources.de.json
```diff
@@ -1,67 +1,70 @@
...
+	"Measured Distance" : "gemessene Distanz",
+	"Switch on to take measurements" : "Auf An drücken, um Distanz zu messen",
+	"Slice Tool" : "Zerschneidungs-Werkzeug",
+	"Axis": "Achse",
+  "Position": "Position",
+  "Select": "Auswählen",
+  "Step": "Schritt",
+  "Tour": "Tour",
+  "of": "von"
+	"Remove tag" : "Tag entfernen",
```
carschno: There are some whitespaces added, presumably not intentional. Could you undo them?

The whitespaces before the `:` were removed in https://github.com/Smithsonian/dpo-voyager/pull/443

assets/language/string.resources.nl.json
```diff
@@ -1,67 +1,70 @@
...
+    "Key" : "Hoofd",
+    "Fill #1" : "Vul #1",
+    "Fill #2" : "Vul #2",
+    "Rim" : "Rand",
+    "Tape Tool" : "Meetgereedschap",
+    "Measured Distance" : "Gemeten afstand",
+    "Switch on to take measurements" : "Schakel in om te meten",
+    "Slice Tool" : "Snijgereedschap",
+    "Axis" : "As",
+    "Position" : "Positie",
```
carschno: Like above, looks like this in undoing the changes from https://github.com/Smithsonian/dpo-voyager/pull/443.

source/client/ui/properties/PropertyTags.ts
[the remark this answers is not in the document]
carschno: I have thought of it now because it would remove the need for parsing the comma-separated lists upon (de-)serialization, but it might be out of scope for this one indeed.

carschno (changes requested):
This looks good to me now; only the whitespace changes need to be un-done (see inline comments).

assets/language/string.resources.de.json
carschno: It looks like merging `master` should actually have done the trick. Something seems to have gone wrong there; perhaps not the most recent `master` branch was merged?

ndejardin425: Corrected

assets/language/string.resources.nl.json
ndejardin425: corrected

carschno (approved):
All looks good to me now, thanks!

gjcope:
Thanks for taking a look @carschno . I think it looks pretty good. My only reservation is how the new property binds the tag list. It definitely works, but is a bit inconsistent from the rest of the properties as it internally relies on a path to a specific element rather than passing in what is needed. It makes it more work to update in the future if we change the tag cloud representation or where it's stored.

carschno:
I agree it should be consistent. @gjcope could you point us to a specific example, given your familiarity with the code base? I might be able to propose an example for this case.

ndejardin425:
> Thanks for taking a look @carschno . I think it looks pretty good. My only reservation is how the new property binds the tag list. It definitely works, but is a bit inconsistent from the rest of the properties as it internally relies on a path to a specific element rather than passing in what is needed. It makes it more work to update in the future if we change the tag cloud representation or where it's stored.

I see your point, I am not a big fan of prop drilling (as it was the other solution) but indeed resolving the tagcloud from a internal path is not a good idea, even more in an opensource project.

So now, CVTask.ts gets the tags cloud from this.activeDocument?.setup.viewer.outs.tagCloud.value then each speciffy Task (AnnotationsTaskView, ToursTaskView, ...) passes the tag cloud to property view as prop, which then passes it to property tags.

Is it possible to do it more cleaner ?

gjcope:
Thanks @carschno and @ndejardin425 for taking a look.

> could you point us to a specific example, given your familiarity with the code base?

I thought some more about this and realized that this is quite similar to some of our regular 'options' properties with dropdowns, like audio and article in annotations and overlay maps for the models. With those properties we update the listen for events that indicate option changes and then update the relevant option choices in the specific component with that property.

This keeps it componentized and event driven like most of the architecture.

I put together a quick and dirty example of just the annotation tags here (https://github.com/Smithsonian/dpo-voyager/tree/dev-tags-update). I'm sure it could be cleaner but it seems to work. It's a little bit more code but I think more consistent and flexible for the future.

Any thoughts/concerns?

gjcope:
On further thought we would probably want the option update handler on CVViewer but I think it will be pretty clean.

ndejardin425:
that seems to be a better way, I've tried to mimic what you've done for the example and put a max of code in CVViewer and remove the prop-drilling.

gjcope:
@ndejardin425 Thanks, this is looking really great!

It works as-is, but I think is actually even simpler. Since we are updating the base property type schema in CVViewer, if we just update one property (like activeTags) it should be automatically available for all properties of that type without pushing to individual components. I gave it a quick try and seems to work. The only thing I noticed, which may be unrelated, is that a tag still remains in the available tag list after being deleted until you move to a new property.

gjcope:
> The only thing I noticed, which may be unrelated, is that a tag still remains in the available tag list after being deleted

So this was definitely related, and due to my simplified test only updating one property attribute. Even though the schema updates are redundant, turns out we definitely need the 'change' event triggered on each property as you implemented to handle deletion. Unfortunate but something to optimize down the line!

So we are all good as-is and I have merged to rc-61: https://github.com/Smithsonian/dpo-voyager/tree/rc-61

PR #447, opened 2026-03-12
Replace Pin markers with configurable 3D MeasurementMarker in tape tool

ndejardin425:
## Summary                                                                                                                                                                                                                                                                                                        
  - Replace the single "pin" marker style with multiple non-invasive 3D marker options for measurement points                                                                                                                                                                                                       
  - Add a new `MeasurementMarker` class hierarchy with 5 marker styles: Sphere, Ring, Crosshair, Disc, and Pin                                                                                                                                                                                                      
  - Marker style is now configurable via the Tape Tool interface                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                                                                                                                        
  - **New file `MeasurementMarker.ts`**: Implements abstract base class and 5 marker styles designed for museum contexts where pin-style markers suggesting surface penetration are inappropriate                                                                                                                   
  - **Updated `CVTape.ts`**: Replaced `Pin` with `MeasurementMarker`, added `markerStyle` property with dynamic style switching support                                                                                                                                                                             
  - **Updated `setup.ts` schema**: Added `TMarkerStyle` type and optional `markerStyle` property to `ITape` interface                                                                                                                                                                                               
  - **Added translations**: New strings for marker style names in 7 languages (EN, FR, DE, ES, IT, JA, NL)                                                                                                                                                                                                          
  - **Removed `Pin.ts`**: Functionality moved to `PinMarker` class within the new marker system                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                    
  ## Marker Styles                                                                                                                                                                                                                                                                                                  
  | Style | Description |                                                                                                                                                                                                                                                                                           
  |-------|-------------|                                                                                                                                                                                                                                                                                           
  | Sphere | Semi-transparent sphere floating above surface (recommended default) |                                                                                                                                                                                                                                 
  | Ring | Flat torus ring sitting on the surface |                                                                                                                                                                                                                                                                 
  | Crosshair | Targeting reticle with crossed lines |                                                                                                                                                                                                                                                              
  | Disc | Flat circular disc on the surface |                                                                                                                                                                                                                                                                      
  | Pin | Legacy pin style (preserved for compatibility) |                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                    
  ## Test Plan                                                                                                                                                                                                                                                                                                      
  - [ ] Verify each marker style renders correctly in the 3D viewer                                                                                                                                                                                                                                                 
  - [ ] Test dynamic style switching preserves marker position and orientation                                                                                                                                                                                                                                      
  - [ ] Confirm measurements work correctly with all marker styles                                                                                                                                                                                                                                                  
  - [ ] Check translations display properly in all supported languages


ndejardin425:
@gjcope did you had time to look it is ? 

To give you the full picture, even if it's virtual, some of our users were not keen on placing pins on old objects. It's purely psychological, but I can understand it.

gjcope:
Hey @ndejardin425 , apologies for the delay.

I had a look and the only thing that concerns me is that it adds a decent amount of code to the frontend Explorer package that will likely not get used on our end.

We haven't gotten any similar feedback from our curators, but I can see the argument. When time allows I'd like to demo the options you put together and maybe pick an alternate to include based on consensus. There's also something to be said for consistency across experiences. Do foresee needing all of these options for your use cases?

ndejardin425:
> Hey @ndejardin425 , apologies for the delay.
> 
> I had a look and the only thing that concerns me is that it adds a decent amount of code to the frontend Explorer package that will likely not get used on our end.
> 
> We haven't gotten any similar feedback from our curators, but I can see the argument. When time allows I'd like to demo the options you put together and maybe pick an alternate to include based on consensus. There's also something to be said for consistency across experiences. Do foresee needing all of these options for your use cases?

No, we won't need all the options, just a "non-invasive" alternative to the pin marker

gjcope:
@ndejardin425 Is there a reason this choice is Explorer-side? It seems like it would be a Story-side configuration option that is serialized in the scene file. If you have a good use case for this though, I am open.

ndejardin425:
for me, it should depend more on the configuration of the explorer than an option in the scene.

gjcope:
> for me, it should depend more on the configuration of the explorer than an option in the scene.

So you anticipate end users needing to change the pin style? I'm not sure our users would make use of that. I would think that organizationally if you are opposed to pins, you would change the style used for your scenes. I'm not sure I see the value of the added front-end UI. Are you getting requests for end user configurability (outside of just having a marker other than a pin)?

ndejardin425:
> > for me, it should depend more on the configuration of the explorer than an option in the scene.
> 
> So you anticipate end users needing to change the pin style? I'm not sure our users would make use of that. I would think that organizationally if you are opposed to pins, you would change the style used for your scenes. I'm not sure I see the value of the added front-end UI. Are you getting requests for end user configurability (outside of just having a marker other than a pin)?

I think it would better to change the style used in Voyager components configuration than in every scene we will create or use. So the measure marker would be another one than a pin for all our users, no configuration based on each user.
For our users, the requirement is just having a marker other than a pin.

gjcope:
It seems like what you are looking for then is the ability to configure a different pin default for the viewer. I don't think we need pin options on the end user side unless there is a compelling motivation or use case for it. Would an API attribute work for you?

ndejardin425:
> It seems like what you are looking for then is the ability to configure a different pin default for the viewer. I don't think we need pin options on the end user side unless there is a compelling motivation or use case for it. Would an API attribute work for you?

Yep, it will work for us.

gjcope:
Hi @ndejardin425 , apologies for the long delay in getting back to this. Will [this branch](https://github.com/Smithsonian/dpo-voyager/tree/dev-markers) work for you? 

It's basically just a pared back version of your PR but maintains the structure so you can add in your other pin options on a fork if needed. It also includes the API attribute to set marker style as discussed. Let me know what you think.

ndejardin425:
Hi @gjcope, seems nice to me. Thanks !

gjcope:
Great! Merged to [rc-64](https://github.com/Smithsonian/dpo-voyager/tree/rc-64)

PR #464, opened 2026-05-12
Crash when uploading images to articles if articles folder not yet in asset tree

ndejardin425:
## Problem

   When inserting an image into an article via the TinyMCE editor, `uploadFile` receives `undefined` as the `folder` argument and crashes:

   ```
   Uncaught TypeError: Cannot read properties of undefined (reading 'info')
       at CVMediaManager.uploadFile
   ```

   **Root cause:** `getAssetByPath("articles/")` looks up the folder entry in `_assetsByPath`, which is populated asynchronously during `refresh()`. If the asset tree has not finished loading when the upload is triggered, the lookup returns `undefined` even though the `articles/` folder exists on the
   server.

### DIFF
--- a/source/client/components/CVMediaManager.ts
+++ b/source/client/components/CVMediaManager.ts
@@ -96,10 +96,10 @@ export default class CVMediaManager extends CAssetManager
         return this.assetManager.getAssetUrl(resolvePathname(uri, this.rootUrl));
     }
 
-    uploadFile(name: string, blob: Blob, folder: IAssetEntry): Promise<any>
+    uploadFile(name: string, blob: Blob, folder?: IAssetEntry): Promise<any>
     {
         const filename = decodeURI(name);
-        const filepath = folder.info.path.length > 1 ? folder.info.path + filename : filename;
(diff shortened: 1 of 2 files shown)


gjcope:
Thanks @ndejardin425 - do you have a good way to reproduce this race condition? Looks like a valuable fix, I just want to be able to test it out.

ndejardin425:
> Thanks @ndejardin425 - do you have a good way to reproduce this race condition? Looks like a valuable fix, I just want to be able to test it out.

I've got the problem each time I tried to upload and add an image to an article. Maybe it is related to how I use Voyager Story, because I could not reproduce it here -> https://3d.si.edu/voyager-story-standalone

gjcope:
Ok. Well it doesn't seem to cause any issues as it just shifts the hard-coded folder name. I can't recall why it was done with a lookup to begin with but if issues pop up in future testing we will revisit. Since you have encountered the issue it's likely others have as well so merged to [rc-62](https://github.com/Smithsonian/dpo-voyager/tree/rc-62) for further testing.

PR #465, opened 2026-05-13
Emit Scene Loading events

sdumetz:
the current `model-load` isn't appropriate to get a sense of "ready for interaction" for multi-model scenes. Even worse with LOD-enabled scenes where we can't even rely on the total count of "model-load" events.

So this PR introduces `scene-loaded`: a new public-facing event that is tied to the application's loading state (`scene-loaded` fires when the `<sv-spinner>` disappears).

I'd expect the computation of "initialLoad" to be more reliable than the previous tracking through `CVAssetManager` in the long run.

While considering the issue, a few other events appeared of interest for implementors of the Voyager API: `scene-loading`, to track document changes (tied to `<sv-spinner>` reappearance) and `assets-loading` / `assets-idle` to track "non-blocking loading", allowing us to signal when the scene has truly reached "idle" state.


gjcope:
Hi @sdumetz , thanks for the PR. Some comments below.

1. Do you have any use cases or usage patterns to illustrate the issues you describe? I'll admit we don't use the model-load event much, but 'initialLoad' has been very stable for us in all use cases to signal ready for interaction.
2. I think there is a mismatch in 'scene loaded' expectations. We consider thumb models a 'loading' state rather than a 'loaded' state. While users *can* interact with thumbnail models, we don't want to support the misunderstanding that this is the final state for users that may have low bandwidth.
3. Similarly to above, the mismatch means there can be a long gap between when the loading spinner stops and the interaction prompt begins if there is a large high-res model or low bandwidth. You can simulate this by using in-browser throttling.
4. The spinner starts on load in standalone mode.

sdumetz:
The loader no longer misbehaves in meaningful ways right now. But it has repeatedly in the past  where it stopped too soon or [flickered endlessly](https://archeologie.ecorpus.eu/ui/scenes/Chauvet/view) (LOD scene using v0.59 - it's not immutable so it'll be fixed next time we update the service). I do think "all models have a loaded representation" is a better heuristic than "loader has reached idle state".

The core of our need is to make external API users be able to listen to the loading state in a meaningful way. Our opinion that the spinner should be tied to "scene-loaded" (ie: reached interactive state) is definitely up for discussion.

If I was to design the system tailored entirely to our needs, I'd put a "blocking" loader until "scene-loaded", and change it to an unobtrusive loading indicator tied to "assets-idle". In applications embedding voyager we currently hide the loader using CSS and hook to the app's internal state to do just that. But I'm not sure how well such a design would serve your needs?

On the other hand, there is nothing inherently wrong with considering the scene as loaded when it first reach "idle" state. FWIW that's generally the default  "loaded" state detector for headless browsers tooling. It's just that it's not a great fit for us : our non-LOD scenes generally have only one derivatives. The other ones are [seldom used](https://github.com/Smithsonian/dpo-voyager/issues/254) anyways so we don't bother with them. So the only case where thumbs are loaded is when using LOD-enabled scenes where idle state might be a long way off, especially since the loader don't block interaction: You may delay reaching idle state indefinitely just by looking around even though the scene is perfectly usable as soon as all thumbs are loaded.

I didn't test standalone mode. I'll need to look into that as well as test the system on more various scenes as soon as we agree on what "the system" needs to be.

gjcope:
> I do think "all models have a loaded representation" is a better heuristic than "loader has reached idle state".

I agree with a slight modification - "All models have a loaded representation of greater quality than Thumb". This was the motivation behind the previous refactor and seems to work well for us.

When we generate Thumb derivatives, they are truly thumbnails. Less than 20k polys and 512 textures. Signaling to the user that the scene is "ready" in this state is misleading, especially if it will be a significant amount of time for the true res asset to stream in. 

> If I was to design the system tailored entirely to our needs, I'd put a "blocking" loader until "scene-loaded", and change it to an unobtrusive loading indicator tied to "assets-idle".

If I understand correctly, this seems like static loading. I'm not opposed to it (many other viewers do it) but I just don't see the down side to allowing interaction during load from a user experience perspective. User testing shows that there is an expectation of immediate interactivity, especially with younger users.

sdumetz:
> All models have a loaded representation of greater quality than Thumb"
Not possible. Loading the [Arc de Triomphe](https://3d.humanities.science/CND3D/7778438.o.2026) fully in anything higher than thumb can't be done without a (high end) discrete GPU.

For some scenes we even have to use lower presets for thumbs. IIRC The _Grotte Chauvet_ scene uses a **128x128** diffuse per tile for its thumb quality (number of vertices is generally irrelevant). If we used anything larger we'd run out of VRAM.

> I just don't see the down side to allowing interaction during load from a user experience perspective

Unless LOD mode is enabled AND we use "idle network" as the signal to hide the loader. Then it is possible (and quite easy in low-bandwidth situations) to keep the loader shown forever by just making small viewpoint changes.

Separately, there is nothing wrong with either being able to move during loading or waiting for "network idle" to hide the loader. I'd even say both are quite desirable. But that just can't work when combined this way.

What I was refering to as a "blocking loader" isn't something that would block interaction, but something that would block 'part of) the view, ie. the current loader. 

To illustrate, in one-model scenes, we'd have the current loader when fetching the document and the thumbnail:
initial

Then we'd transition to an unobtrusive (style to be adjusted of course) loader that indicates more content is still being fetched :
loader

This loader would be tied to the `assets-loading` / `assets-idle` events so it'd be shown every time we fetch another resource. If done properly it would also help communicate when we are in the process of loading an overlay, which AFAIK is currently not done at all and can be quite frustrating when they get a bit on the heavy side.

Not to say that's the best or only way to communicate loading state. Just that we have to find a system that works for all scenes, and being able to communicate non-initial loading state would be great as we are regularly (https://github.com/Smithsonian/dpo-voyager/discussions/298, https://github.com/Smithsonian/dpo-voyager/pull/324, https://github.com/Smithsonian/dpo-voyager/pull/427) adding such new resources.


This leaves open the question of **when** the loader should transition from "initial" to "busy". I understand your concern around the thumb quality situation but I can't think of something better that would work everywhere. Maybe you have something in mind?

gjcope:
> > All models have a loaded representation of greater quality than Thumb"
> > Not possible. Loading the [Arc de Triomphe](https://3d.humanities.science/CND3D/7778438.o.2026) fully in anything higher than thumb can't be done without a (high end) discrete GPU.
> 
> For some scenes we even have to use lower presets for thumbs. IIRC The _Grotte Chauvet_ scene uses a **128x128** diffuse per tile for its thumb quality (number of vertices is generally irrelevant). If we used anything larger we'd run out of VRAM.

That's fair. Then for those use cases I would agree "all models have a loaded representation" makes sense, but for nearly all of our current use cases it does not. 

> Unless LOD mode is enabled AND we use "idle network" as the signal to hide the loader. Then it is possible (and quite easy in low-bandwidth situations) to keep the loader shown forever by just making small viewpoint changes.
> 

This would be true with your 'non-blocking' proposal too, no? LOD seems like a special case for loading indication.

> What I was refering to as a "blocking loader" isn't something that would block interaction, but something that would block 'part of) the view, ie. the current loader.
> 

Understood. I still think the fundamental issue is a difference in view point on when a scene is ready for interrogation. When it is not, I think the "blocking loader" is appropriate.

> This loader would be tied to the `assets-loading` / `assets-idle` events so it'd be shown every time we fetch another resource. If done properly it would also help communicate when we are in the process of loading an overlay, which AFAIK is currently not done at all and can be quite frustrating when they get a bit on the heavy side.
> 

I'll discuss the idea more internally, but this is essentially how the very first loading indicator operated. users found it disruptive as you don't really want "every" resource to prompt an indicator. You get a significant amount of flickering for quickly loading assets. You would also fall right back into your LOD example above where you have an indicator always visible.

> This leaves open the question of **when** the loader should transition from "initial" to "busy". I understand your concern around the thumb quality situation but I can't think of something better that would work everywhere. 

Agreed. For non-LOD scenes (99% of our current repo) we would consider "initial" finished when all assets in the scene have the "desired" derivative quality loaded. This could be the default High or something lower as indicated by the scene file or other parameters (or in the future even bandwidth, resources, etc.). For LOD I would agree with  "all models have a loaded representation".

For additional context, we currently use the GA hooks in Voyager to track many different stats including load time. This is tied directly to that "initial" transition. If this were only tracking the load time to "thumb", these stats would be much less helpful in guiding decisions like package size limits, CDN usage, etc.

sdumetz:
> You get a significant amount of flickering for quickly loading assets. You would also fall right back into your LOD example above where you have an indicator always visible.

Not that much of a problem when the "busy" indicator is tucked away in some corner and not too distractive. We found a fade-in-out animation gracefully hides away the flickering for the most part.

With that said, what would you say to a PR that would keep the current "initialLoad" mechanism but would expose the two public events proposed before, with only the first `"assets-idle"` event mapping cleanly to a change in the loading indicator?

On our side, we'd hide the loader and wire those events as needed. That's not ideal, but that's a way forward.

My opinion however remains that the current state is not satisfactory at all. It looks like I'm not [the only one](https://3dviewer.digiarc.mt/model-browser.html?selected=Benghisa_Voyager) thinking that, even if I'm the only one requesting changes. And their implementation is quite obviously broken too. That's not their fault; I have yet to find a way to make a not-broken loading indicator on top of voyager with the current API even including the non-public internal events you may hook to through the `system` property of the explorer application.

gjcope:
@sdumetz Coming back to this and hoping to find a path forward.

> Not that much of a problem when the "busy" indicator is tucked away in some corner and not too distractive. We found a fade-in-out animation gracefully hides away the flickering for the most part.

This is subjective so something we would user test if needed.

> With that said, what would you say to a PR that would keep the current "initialLoad" mechanism but would expose the two public events proposed before, with only the first "assets-idle" event mapping cleanly to a change in the loading indicator?

The main evidence I've seen so far of the current "initialLoad" mechanism not working is for your LOD use cases. This would be an easy fix. But that being said, I'm not opposed at all to your re-implementation - it's purely the timing. The current proposed solution makes sense to me from your use case perspective but would break quite a bit of our loading-based functionality as described previously (interaction prompt, GA loading events, standalone ingestion) and change our user experience.

Happy to include the public events as we want the component to be as flexible as possible, but the important bit from my perspective is that they are aligned appropriately. It seems we still differ on what qualifies as idle or ready for interaction.

Since there is already the concept of the desired derivative quality (CVViewer.ins.quality) maybe we are in need of two scene load events, one to indicate an asset (any quality) has loaded for each model and one to indicate the requested initial quality has loaded. In your case these would most often fire at the same time but in our use cases they would not.

As far as the asset specific events, I don't see any issues with providing those. The concern would be if we affecting the internal load state driving things like the interaction prompt, etc. described previously.

sdumetz:
Hi, thanks for taking the time, that's much appreciated.

> the important bit from my perspective is that they are aligned appropriately. 

definitely. I'd rather delay this than have a system that doesn't align internal and external state properly.

Creating more events tied to _a_ model rather than the global scene state would create more confusion IMO. That's already a problem with the current `'model-load'` event, that is difficult to reconcile with how the scene _looks_ when the scene graph is non-trivial.

I'm thinking maybe:

 - `scene-load`: new event, when all models have _a_ representation
 - `scene-idle`: tied to CVAssetManager's busy state, with a parameter indicating "initialLoad" or something like this.

It wouldn't address our LOD endless loading state. But since I've been playing a lot with [Nexus](https://ecorpus.fr-scv.fr/scenes/Chauvet_Nexus/index.html) the last few weeks, I'm more open to the idea that there might be better (or other, at least) ways to tackle our need for dynamic-resolution models, so that might not matter _that much_ in the long run.

gjcope:
> Creating more events tied to _a_ model rather than the global scene state would create more confusion IMO.

Sorry if I wasn't clear, that wasn't what I was proposing. I was just proposing two scene state events based on the conditions described. It seems like that would suit both of our needs. It also seems like what you are proposing above unless I'm misunderstanding. 

It's unclear to me why 'scene-load' wouldn't address LOD endless loading? If one event fires when you have *something* loaded for all visible objects it seems like that would do it.

> But since I've been playing a lot with [Nexus](https://ecorpus.fr-scv.fr/scenes/Chauvet_Nexus/index.html)

I'm getting a 404 but interested to see!

sdumetz:
Oh, sorry about the 404, I forgot to set the permissions. It should work now.

The scene-load event would be 100% right for LOD. It's keeping the "spinner" tied to idle state rather than "initial load" that I was referring to.

But this initial set of changes wouldn't make anything worse and we _could_ get what we need by hiding the loader and providing our own overlay tied to the new event in cases where this is required, so that's definitely an improvement.

I'll try to implement that.

sdumetz:
- [d1a5d29](https://github.com/Smithsonian/dpo-voyager/pull/465/commits/d1a5d29364556890ea41faa81fc992122eedc40d) introduces a minimal fix for initialLoad to trigger as it should for LOD scenes. It's still possible to delay it infinitely by moving in the scene and never allowing it to fully settle, but once it settles, the spinner won't come back.
- [1c7b589](https://github.com/Smithsonian/dpo-voyager/pull/465/commits/1c7b58963c0ec9b78a4ba12b616414203db1e410) introduces `scene-content-load` and `scene-load` events. It also adds a `model-error`event. saving faield state for models was required to properly track loading so I thought we might as well surface it: it's a good complement to `model-load`.

I'm not sure about the names. My reasoning was: `initialLoad` is preexisting so it felt right to map it to `scene-load`. `scene-content-loaded` kinda maps to [DomContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event), though they don't fully overlap.

My first idea was `scene-load` -> `assets-idle` (or `assets-load` to keep the semantic link to `initialLoad`), but the public API has no concept of "assets" and the spinner certainly _feels_ scene-related.

Interested to have your thoughts on this.

gjcope:
Thanks for the update! I think it looks great and functions well.

`scene-content-load` does seem a little awkward, but I can't think of anything significantly better. I think you do a good job of explaining it in the docs and the tie-in to DomContentLoaded should be helpful for those who are familiar with it.

The only small difference I'm seeing with current behavior is a gap in spinner visibility between the loading of the template and main scene. But that's minor.

I also wanted to reiterate that I totally see where there could be value for asset-level events, your earlier overlay example being a good one, but that could be for another time.

gjcope:
Merged with [rc-64](https://github.com/Smithsonian/dpo-voyager/tree/rc-64)

Then realized after the fact, do we need to clear out the _loadedModel references on dispose? Have not tested that yet.

PR #467, opened 2026-05-28
Remove "Reverse" animation type and allow negative animation speed instead

Inao0:
This would have the benefit to work for all styles of animations (Single, PingPong and Loop) instead of having only the "Reverse" type which is actually a "Single reverse".


gjcope:
Makes sense, thanks!

PR #471, opened 2026-06-06
Add preliminary Kintsugi support for Voyager

michaelt919:
- Modified ModelReader to automatically detect and load Kintsugi textures referenced in the glTF userData -- no longer necessary to specify them manually.
-  Modified ShaderExtension with the required shader code injections for Kintsugi models to appear correctly.  A custom implementation of RE_Direct and RE_IndirectSpecular have been provided that strip out much of the physically-based rendering in order to match Kintsugi's optimization model.  The specular color is also now correctly overridden even for functions that use specularColorBlended.

### DIFF
--- a/source/client/io/ModelReader.ts
+++ b/source/client/io/ModelReader.ts
@@ -107,6 +107,38 @@ export default class ModelReader
         return this.loadModel(url, {signal})
         .then(data=>this.gltfLoader.parseAsync(data, resourcePath))
         .then(gltf=> {
+            // Check for Kintsugi materials and extract them automatically from the glTF
+            gltf.scene.traverse(object => {
+                const material = object["material"] as MeshStandardMaterial
+
+                if (material) {
+                    // Use Kintsugi diffuse instead of the albedo map (which incorporates both diffuse and specular)
+                    var kintsugiDiffuse = material.userData["diffuseTexture"]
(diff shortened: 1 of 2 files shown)


gjcope:
Thanks @michaelt919 this is great! It will take a bit to test, but it all looks reasonable.

Re: Kintsugi diffuse texture - does this mean that a glTF file produced with Kintusgi will have an albedo _and_ diffuse texture? Does this diffuse incorporate any other shading components besides specular (occlusion, etc.)? Does this affect dynamic lighting scenarios?

michaelt919:
Yes, the Kintsugi glTF will have a PBR albedo as as well as a diffuse texture -- and they are not the same.  The PBR albedo includes both diffuse content as well as specular content for metals.  The Kintsugi diffuse is just diffuse.  Kintsugi exports the PBR albedo for compatibility with generic renderers as well as for some global illumination approximations that don't support separate diffuse and specular maps.  However, once you dig in under the hood, Voyager actually separates diffuse and specular fairly cleanly, so the PBR albedo map isn't really needed.

If desired, I could add a feature to Kintsugi to export an "optimized" Voyager output that wouldn't include any unneeded textures.

Occlusion should be packed into the red channel of the ORM map (as with albedo, we don't really need metallicity but the occlusion and roughness are required).

PR #476, opened 2026-06-26
Raking Light

### LINKED ISSUE #31: Raking Light
> original article describing the 3DHOP viewer - https://iris.unimore.it/bitstream/11380/1170215/1/3DHOP%203D%20Heritage%20Online%20Presenter.pdf. Particularly the section "4.3. Interaction components" - this is where the trackball feature is described in coding terms.

- Discussion: https://github.com/Smithsonian/dpo-voyager/discussions/475
- PR: https://github.com/Smithsonian/dpo-voyager/pull/476
- [Preview](https://dpo-voyager-git-feature-rakinglight-carstens-projects-7b21f278.vercel.app/voyager-story-dev.html?document=https://yhp1co7ikmvyqrk7.public.blob.vercel-storage.com/files/models/tusk/tusk.svx.json)

carschno:
Add a [raking light](https://conservation-wiki.com/wiki/Raking_light). 
Closes https://github.com/Smithsonian/dpo-voyager/discussions/475.

[Preview](https://dpo-voyager-git-feature-rakinglight-carstens-projects-7b21f278.vercel.app/voyager-story-dev.html?document=https://yhp1co7ikmvyqrk7.public.blob.vercel-storage.com/files/models/tusk/tusk.svx.json)

# Changes

- ~~Add `CVRakingLight`. Inherits from `CVDirectionalLight`, and adds `azimuth` and `elevation` parameters.~~
- Add `Raking Light` switch to Directional Light controls, enabling the trackball widget for control.
- Add trackball widget for raking light control

# TODO

- [X] The UI shown in the screenshot needs refinement due to the large space it takes and the redundancy (hemisphere *and* sliders).

# Screenshot

image

The hemisphere is used for easier control of the elevation and azimuth parameters (see https://github.com/Smithsonian/dpo-voyager/discussions/475#discussioncomment-17420652). ~~Alternatively, both parameters can be configured using the respective sliders.~~


gjcope:
@carschno Thanks for the PR - very nice! Functionally, with the exception of the UI issues you've already noted, I think it works great. From an implementation standpoint I have a couple of thoughts.

- Do we need a new light type? I think this would work well for directional lights in general (and potentially others, though that could get confusing) 
- I'm curious on the performance implications of a 2D solution vs 3D. 3D is likely less code in this context but I would guess your implementation is more performant. Are you aware of any reference implementations?
- Seems like this could also be useful in Story for positioning lights, but I wonder about confusion if it's only available for directional. Probably still value in that.

kgschoueri:
@gjcope and @carschno I can try to enter the conversation about UI/Implementation. 

- I agree with Jaime that in the end, it is the same as a directional light. I supposed what we aim for is a way that an end-user can intuitively adjust the direction via the trackball itself (I don't love the sliders for azimuth and elevation as the way to do that). Is this what you meant Jaime by 2D solution vs 3D solution? 

- functionally, I think we need to rethink the location of the trackball. From the screenshot, I assume you can only adjust the location of the light from the tools menu but I find that annoying because then you can't simutaneously see the effect the changing light has on the object as it is largely obscured by the overlay of the tool. Maybe it can be more discrete, both in terms of size, design and location within the viewer. (also perhaps the 3D example that Carsten is using is not the ideal use case for this tool, as it was for the sunlight tool). 

- With that in mind, I also assume the UI/UX will change with the front-end redesign so maybe we can wait for the more aesthetic aspects of the tool and focus on the logic of the light itself. I can look into examples to get ideas on that. 

- A final thought is more about the amount of 3D literacy required to fully understand the tool and what kind of 3D models it is suitable for, e.g., higher-resolution geometric mesh and a knowledge of what on the 3D model has been modified by human or computer interventions (e.g., smoothing to make a model look aesthetically 'clean'). I'm aware this requires a wider awareness of 3D limitations but perhaps we can start thinking about how to make this literacy more accessible via the interface or on the web platform. Anyway, this was something I realized in my work with 3D interfaces, that tools are not always transparent as to their utility in different contexts or their limitations depending on the type of the 3D model.

carschno:
Closes https://github.com/Dynamic3D-Project/dpo-voyager/issues/31

gjcope:
> @gjcope and @carschno I can try to enter the conversation about UI/Implementation.
> 
> * I agree with Jaime that in the end, it is the same as a directional light. I supposed what we aim for is a way that an end-user can intuitively adjust the direction via the trackball itself (I don't love the sliders for azimuth and elevation as the way to do that). Is this what you meant Jaime by 2D solution vs 3D solution?

Agree, I think just the ball would do the job. The 2D/3D comment was more of a technical one. @carschno 's current implementation draws the 'trackball' and  highlight in 2D instead of using hemisphere geometry and a 3D light. I think it works well but was just curious about the performance differences

> *  Maybe it can be more discrete, both in terms of size, design and location within the viewer. 

Agree and I think covered by Carsten's TODO

> * With that in mind, I also assume the UI/UX will change with the front-end redesign so maybe we can wait for the more aesthetic aspects of the tool and focus on the logic of the light itself. I can look into examples to get ideas on that.

I don't think things will change dramatically as far as the space we have for tools, but agree it makes sense to focus on the functionality first.

> * A final thought is more about the amount of 3D literacy required to fully understand the tool and what kind of 3D models it is suitable for, e.g., higher-resolution geometric mesh and a knowledge of what on the 3D model has been modified by human or computer interventions (e.g., smoothing to make a model look aesthetically 'clean'). I'm aware this requires a wider awareness of 3D limitations but perhaps we can start thinking about how to make this literacy more accessible via the interface or on the web platform. Anyway, this was something I realized in my work with 3D interfaces, that tools are not always transparent as to their utility in different contexts or their limitations depending on the type of the 3D model.

There's definitely bias here, but I personally have always found the hemisphere-with-highlight control for lights to be the most intuitive as it literally shows the light changing on an object, whether or not the complexity of the scene object reflects it as transparently. But I also think some of the onus is on the scene author to use lighting that is appropriate for the content. I.e. if this content would benefit from raking light, there is a directional light available in an appropriate place to make that happen.

carschno:
I have reworked the UI/UX, removing the dedicated Raking Light type. Instead, the Directional Light controls now provide a Raking Light switch to enable the trackball control for azimuth and elevation. The screenshot is updated accordingly.
I have also added a Preview link in this issue's description above. Please check that preview to play with the current draft.

gjcope:
@carschno The reworked UI is great! We may be even be able to get away with not having the raking on/off button, but that might depend on my question below.

From a user experience standpoint it seems a bit off. But this may be a mismatch between the specific requirements for a "raking" light vs just intuitive light direction control.

My initial expectation as a user is that the hemisphere represents the light from my perspective. I.e. if I see a highlight from the left of the UI element I would expect the light to be shining on the left of the object from my current view.

Playing around with it, it seems the hemisphere is static and represents the top-half of the scene space. If this is true, I would then expect if the highlight showed all the way on the right side of the UI that we would only get light on the right side of the object. But that doesn't seem to quite be the case. See the pic below with an axis-aligned cube viewed from the front.

If we can get some more clarity on the requirements/expectations for the raking light from @kgschoueri and team that would help to inform the best path forward.

image

carschno:
> My initial expectation as a user is that the hemisphere represents the light from my perspective. I.e. if I see a highlight from the left of the UI element I would expect the light to be shining on the left of the object from my current view.
> 
> Playing around with it, it seems the hemisphere is static and represents the top-half of the scene space. If this is true, I would then expect if the highlight showed all the way on the right side of the UI that we would only get light on the right side of the object. But that doesn't seem to quite be the case. See the pic below with an axis-aligned cube viewed from the front.

I see how this is not intuitive, but coming from the definition of a raking light, the use should envision a perspective from the top, which is how the trackball should be envisioned. If the position marker in the trackball is in the center, it means that the raking light comes straight from above. It corresponds to both azimuth and elevation being 90 degrees.
I understand that this is not in line with what you see, but it is in the screenshot below (from the [preview](https://dpo-voyager-git-feature-rakinglight-carstens-projects-7b21f278.vercel.app/voyager-story-dev.html?document=https://yhp1co7ikmvyqrk7.public.blob.vercel-storage.com/files/models/tusk/tusk.svx.json)), which shows the shadows as intended with a raking light.

In your example with the box, what would be the expected behaviour? Would you expect the front to be completely dark?

image

Do you suspect an implementation failure here or rather a counter-intuitive UX? In case of the latter, I agree that the raking light is not quite intuitive, but I am not sure how to make it more user-friendly. Perhaps @kgschoueri can explore this further indeed.

BTW, the trackball is limited to positioning the light to the upper hemisphere of the model indeed. By manually setting the `elevation` to a negative value in the light settings, however, you can actually position it into the lower hemisphere. If this a feature that should be controllable by a user, ie. a raking light from below, we could think about adding a switch that inverts the trackball so that it represents the lower hemisphere of the model.

gjcope:
> coming from the definition of a raking light, the use should envision a perspective from the top, which is how the trackball should be envisioned.

Can you point me to this definition? In the wiki you linked above, I see "Raking light is an imaging technique where a single light source is positioned at a sharp angle to the object" and "Illumination in raking light should be done from the top and from the left to avoid perceptual biases". With the current implementation there are a number of viewpoints where you would not be able to achieve this. If the hemisphere represented your current view, you could always get a light source at a sharp angle to the object by dragging around the outer edge of the UI element.

> In your example with the box, what would be the expected behaviour? Would you expect the front to be completely dark?

Yes, exactly. In fact if I move around the outer edge of the UI widget, the front face is dark 50% of the time (as expected), but it's as if the hemisphere is rotated. Positioning the light at the "front" does not seem to translate to the scene. See pic below

image

kgschoueri:
>  In the wiki you linked above, I see "Raking light is an imaging technique where a single light source is positioned at a sharp angle to the object" and "Illumination in raking light should be done from the top and from the left to avoid perceptual biases". With the current implementation there are a number of viewpoints where you would not be able to achieve this.

I'm not an expert on this technique, but perhaps what is happening is that we are getting hung up on the "raking light" as a term for the tool. From my understanding, raking light is a specific photographic technique that has it's own protocols. From my perspective, what we are trying to achieve with the tool is a kind of soft Virtual RTI so that geometric surface features on a 3D model can be better percieved. Susan calls it a flashlight and I see the Dynamic Collections calls it simply 'Lighting Direction'. However, indeed we will need to consult stakeholders who use this technique/tool more regularly to see if our direction for the tool indeed is meeting their expectations. 

From my experience, the tool operates as I would expect. After testing a bit more, I noticed what Jaime said about the hemisphere limit of the light. I think also the UI/UX can be improved once we get more feedback, it was not immediately intuitive to me that the light was movable per light selected. I would have assumed it was some hidden light that gets turned on just for the 'flashlight' experience. This might help with UI by predefining the light type, intensity and color so all the user has to do is enable and move around using the trackball.

carschno:
I think we are getting closer to the core of it. I agree that Raking Light might not be the best term to use, given that this implementation practically only provides another control for positioning a (directional) light.

So I wonder:
1. Can we find a name that focuses on the fact that it enables the trackball, along the lines 'Positioning Trackball (from top)'? What reflects the actual purpose best? Flashlight does not seem entirely clear to me either at first glance...
2. Should we provide that option for every light type; or, inversely, is there a reason to keep it restricted to the Directional Lights?
3. Should we add another parameter to let the trackball represent the *lower* hemisphere of the scene, instead of the top (ie inverse the elevation value)?

gjcope:
I'm still stuck on why it needs to be the top hemisphere. If it is supposed to mimic a "flashlight" or "virtual RTI" like @kgschoueri mentioned, it seems like view-dependent control would be ideal. I may still be missing a requirement or something.

PR #483, opened 2026-07-09
Add support for multiple models using the same animation name

Inao0:
The _animMap dictionary in CVActionManager currently uses only the name of the animation as its key. 
Using the string `model.id + animation.name` instead allow to support multiple models using the same animation name. 

To facilitate picking between similar actions on the different models, this pull request also displays the model name in front of the action name in the dropdown for "OnActionEnd" and "OnActionBegin" triggers.

### DIFF
--- a/source/client/components/CVActionManager.ts
+++ b/source/client/components/CVActionManager.ts
@@ -239,7 +239,7 @@ export default class CVActionManager extends Component
         model.object3D.traverse(object => {
             if (object.animations.length > 0) {
                 object.animations.forEach((anim) => {
-                    this._animMap[anim.name] = object;
+                    this._animMap[model.node.id + anim.name] = object;
                 })
             }
         });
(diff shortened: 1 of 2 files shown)


gjcope:
Thanks for the PR @Inao0 ! We will definitely update the animation lookup but there is some concern on our end for the name prefixing in the dropdown. I get the value, but in some cases it could be counterproductive when long action (or model) names are involved as we have seen the necessity for recently with more involved action logic (See pic below). Do you have a compelling example you can share of the value-add for this?

image

Inao0:
I tested the new actions trigger on an older model with actions. In this model we have multiple identical small arrows/particles following a path to highlight the path taken by the electric current. They all follow the same path, but with different time offsets at the export. Until now all these actions had the same name because they were doing the same thing and were only visible in the context of their own model.

This makes it difficult to identify them in the new trigger dropdown :

image

I guess this could also be solved by changing the naming scheme to manually include a reference to the model when necessary.

gjcope:
Got it, thank you for the example. I think for now we will merge the indexing but not the prefix. I do see the value, but would prefer to keep it simple for now and see in regular use if it becomes prohibitive. If it does, please let me know and we will add the prefix as well.

gjcope:
Partially merged to [rc-64](https://github.com/Smithsonian/dpo-voyager/tree/rc-64) as described above

PR #484, opened 2026-07-10
Fix panning speed for non-default fov value (ie !=52°)

Inao0:
After testing on a scene where the fov value was reduced to 16°, we noticed that the panning behavior was way to fast.
This was due to the x and y offset not taking into account the fov values but working well for the default 52° value.


gjcope:
@Inao0 Looks good, thanks! Merged to [rc-64](https://github.com/Smithsonian/dpo-voyager/tree/rc-64)

PR #488, opened 2026-08-03
Feature/video shader

### LINKED ISSUE #34: Moving Images
# Idea

Make it possible to project a custom video on a surface to simulate the screening experience from a video projector

carschno:
# Motivation

Project videos on a surface to simulate (home) cinema scenes.

Closes https://github.com/Dynamic3D-Project/dpo-voyager/issues/34

# UI Changes

- Rename the Audio task to `Multimedia`, and add a `Create Video` button to create video elements
- Action task: add a `PlayVideo` action to select a video element that is projected to a model in the scene

# Code Changes (Overview)

Allow setting a custom video URL as [VideoTexture](https://threejs.org/docs/#VideoTexture)

- `CVMultimediaManager`: abstract class to implement and define common properties and methods of `CVAudioManager` and `CVVideoManager` (new)
- `CVAudioManager`: now extends `CVMultiMediaManager`, remove inherited methods
- `CVVideoManager`: functionality similar to `CVAudioManager`, with video-specific adaptation.
- Rename `CVAudioTask` to `CVMultiMediaTask`
- `CVActionManager`: add `PlayAudio` action
- `CVActionTask`: 
- `CVModel2`: add `applyVideoTextureToMaterials()` and `setupVideoTexture()` to generate and apply `VideoTexture` elements.
- ~~Add new properties in `CVModel2.ts` to specify a `VideoURL` that will be provided by the existing `AssetManager`.~~
- ~~Add a new `EShaderMode.Video` shader to `CVModel2.ts` that uses a video texture as `emissiveMap`~~
- ~~`setup.ts`: add the new `EShaderMode.Video` shader~~

# Preview

- load [home cinema scene with video plane](https://dpo-voyager-git-feature-videoshader-carstens-projects-7b21f278.vercel.app/voyager-story-dev.html?document=https://yhp1co7ikmvyqrk7.public.blob.vercel-storage.com/files/models/home_cinema/scene.svx.json): use `/files/articles/Media/6114_Pathe_News_00_01_01_08.mp4` as example video URL.

https://github.com/user-attachments/assets/5edff032-19b0-4279-823f-050e706e684d


kgschoueri:
@carschno I see that you added in the video as a material that can be toggled on or off via the materials function of the toolbar, but I wonder if the logic of how the video is coordinated can be moved to the Audio and Actions Tasks? This would improve the UX/UI and enable more nuanced control of how the video is being triggered by the author or the end-user. 

I see this as following the current similar workflow for adding in audio mp3 files and then triggering them through a specified action when a certain interaction is made with the 3D model. If that is even possible, then I imagine the Audio Task would have to be renamed, perhaps Media Task or Audio-Visual? This is where the video mp4 is uploaded (instead of under Model Settings Task). Alternatively, perhaps the video texture is uploaded/connected at the moment of creating the action (therefore bypassing the extra step of creating a new audio/video element). 

Screenshot 2026-08-03 at 4 24 13 PM

And then the properties when adding an action to a 3D model would require a new type, i.e., PlayVideo. And, like with PlayAudio, the properties would adjust to the video type for the action. 
Screenshot 2026-08-03 at 4 21 42 PM

And, like with PlayAudio, the properties would adjust to the video type for the action. But have the controls be similar to animation where you can say the style is single, loop, ping pong, etc. 
Screenshot 2026-08-03 at 4 22 18 PM

For me, audio, animation and video texture are all different types of 3D model enhancement that logically should be controlled by the Action Task as it relates to specific 3D models selected in the Navigator node.

gjcope:
Hi @carschno , thanks for the PR!

I think supporting video textures is a great feature, but I have some questions about the implementation. Can you provide some more context for the choices made?

- The Three.js example (https://github.com/mrdoob/three.js/blob/master/examples/webgl_materials_video.html) is pretty simple. Adding a hidden html video element (so the browser can handle the loading natively), creating a video texture and assigning it to a channel, and then playing. Is there a reason for the custom loading logic? We might also consider front loading the asset like we do with audio to avoid interrupting the user experience.
- It's not clear to me what the value of the shader mode option is. Will there be use cases where we would not want to render a texture that is not the video?
- If we need to do our own state management of the video, we should consider a video player element similar to what we do with audio.
- I don't think the Model component is the appropriate place for interaction. I think the ideal solution would be through the action manager or traditional video UI (see above comment) if appropriate.

carschno:
Hi both,
Thanks for your perspectives! I will try to summarise them:

- **UX/UI**: from both of your comments (@kgschoueri , @gjcope), I suggest the following changes:
  - [x] Remove the video shader option from the models/materials settings
  - [x] Rename the *Audio* task to *Multimedia* and generalise it so that the user can create video elements in the same way as audio elements
  - [x] Add a `PlayVideo` option to the *Action* task that includes the video controls (mute, loop, autoplay etc.)
- **Implementation-wise**, I am planning to add a `VideoManager` that works like the current `AudioManager`, rename the `AudioTask` to `MultimediaTask`, and add the `VideoManager` to the new `MultimediaTask`.
- @gjcope I had added the fetching logic because I had some issues loading the videos earlier -- apparently unrelated, so I have removed it.

carschno:
@kgschoueri @gjcope I have implemented the changes, and updated the description, including the screencast. Please feel free to test the preview as linked in the description above!

kgschoueri:
@carschno nice work! I think this works well from a UX perspective! I managed to insert the video and play it with no issues!

carschno:
@gjcope With the recent updates, this feature is ready to merge from our side. If you have more questions or comments about functionality or code, please let us know!

gjcope:
@carschno - A couple of notes/questions:

- To date we've intentionally avoided supporting videos as scene assets (vs streaming) for a number of different reasons - maintenance, package size, accessibility (captions). We advocate towards videos being hosted on streaming services like YouTube and embedded to address these concerns. Would this work here?
- Conceptually I'm not sure how well the audio and video setups fit together as implemented here. For instance, audio elements get created and then can be referenced by/connected to various other parts of the system. With the video, by definition it needs to be attached to a piece of geometry. In your demo, how does the system know to display it on the video plane? I'm not seeing the linkage.
- Small detail, looks like the config panel for video allows to play, but not stop.

kgschoueri:
@gjcope I'll try to answer your questions and @carschno can follow up if necessary. 

- These are valid concerns and ones we didn't consider. For the specific use case, it's important to have a video embedded in the 3D scene to increase immersivity through enhanced dynamic functionality of Voyager. In terms of captions for accessibility, I see the video asset upload also includes a caption path (as with the audio type). We can think about limiting the file size of the upload to ensure a managable package size. Maintenance, I am not sure what can be done regarding that. 
- From my understanding, the video and audio media types are conceptually operating in similar ways. A video is connected to the geometric 3D plane that an author has uploaded separately from the rest of the scene. When the 3D plane model node is selected, the system makes the connection between the previously uploaded video and the geometry using the action task, in exactly the same way as with the play audio type. 
- Interesting. I notice that when I have the video as triggered on click, I can play/pause by clicking the model. But I don't know how it would function for the other trigger types. It was a bit wonky when I tried for the tour step trigger type. Maybe there is a way to add in the video controls as it is with the audio media type. To pause, reverse, fast forward, etc. Would also be useful to know how long the video is.

gjcope:
> * These are valid concerns and ones we didn't consider. For the specific use case, it's important to have a video embedded in the 3D scene to increase immersivity through enhanced dynamic functionality of Voyager. In terms of captions for accessibility, I see the video asset upload also includes a caption path (as with the audio type). We can think about limiting the file size of the upload to ensure a managable package size. Maintenance, I am not sure what can be done regarding that.

The question here was not so much about the video being embedded in the scene, but the video being *stored* with the scene rather than streaming. If the video embedded in the scene can be streamed from a service like YouTube, it would reduce the package size and also give you the accessibility options for free. The UI for the video asset implementation shows the caption path box, but I believe that may just be a remnant of @carschno repurposing the audio element UI. It's not clear to me that it actually works.

> * From my understanding, the video and audio media types are conceptually operating in similar ways. A video is connected to the geometric 3D plane that an author has uploaded separately from the rest of the scene. When the 3D plane model node is selected, the system makes the connection between the previously uploaded video and the geometry using the action task, in exactly the same way as with the play audio type.

It seems to be a little different. The audio setup task, which the video functionality is reusing, creates an element that is independent of any geometry, player, etc. It just exists as an entity that needs to be connected somewhere else to be functional. For instance, by being assigned as the narrative audio or by adding as to the audio slot of an annotation. Looking into the code a little deeper, it looks like the video is only associated with a piece of geometry at the time 'play' is clicked. If I create a video element, select the 'room' geometry, and hit play - the video plays as a texture on the room rather than the video plane. I might be missing the intent, but that doesn't seem like the linkage we would want there.


> * Interesting. I notice that when I have the video as triggered on click, I can play/pause by clicking the model. But I don't know how it would function for the other trigger types. It was a bit wonky when I tried for the tour step trigger type. Maybe there is a way to add in the video controls as it is with the audio media type. To pause, reverse, fast forward, etc. Would also be useful to know how long the video is.

Good point. I hadn't tried out all of those options yet, but noticed that the explicit "Stop" button in the task UI did not seem to do anything.

carschno:
> The question here was not so much about the video being embedded in the scene, but the video being _stored_ with the scene rather than streaming. If the video embedded in the scene can be streamed from a service like YouTube, it would reduce the package size and also give you the accessibility options for free. The UI for the video asset implementation shows the caption path box, but I believe that may just be a remnant of @carschno repurposing the audio element UI. It's not clear to me that it actually works.

- For the streaming/remote URL: using any remote URLs should not be an issue, but I don't think this is practical due to CORS constraints:
```console
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.youtube.com/shorts/Ah_LMYqd2CE. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 302.
```
Anyway, I might be missing the typical scenario, which might imply pointing to arbitrary URLs on the same origin. For YouTube, I do not see how this could work.

The caption box has been a remainder of the audio element UI indeed. I have not been able to test it, but I assumed it would work just like for audio. Perhaps it is better to remove it unless we have a testable use case @kgschoueri ?

> It seems to be a little different. The audio setup task, which the video functionality is reusing, creates an element that is independent of any geometry, player, etc. It just exists as an entity that needs to be connected somewhere else to be functional. For instance, by being assigned as the narrative audio or by adding as to the audio slot of an annotation. Looking into the code a little deeper, it looks like the video is only associated with a piece of geometry at the time 'play' is clicked. If I create a video element, select the 'room' geometry, and hit play - the video plays as a texture on the room rather than the video plane. I might be missing the intent, but that doesn't seem like the linkage we would want there.

The idea from a UX perspective is indeed to have a model for which it makes sense to play to video on, like the video plane. Linking it to the room, for instance, is technically not different. But I would put the responsibility on the user here.
 
> Good point. I hadn't tried out all of those options yet, but noticed that the explicit "Stop" button in the task UI did not seem to do anything.

I had not noticed. Should be an easy fix, but I feel like we should clarify the conceptual UX concerns first.

gjcope:
> The idea from a UX perspective is indeed to have a model for which it makes sense to play to video on, like the video plane. Linking it to the room, for instance, is technically not different. But I would put the responsibility on the user here.

I'm not following the logic here. Right now there is no way that I can tell to "link" the video with the geometry. It just plays on whatever geometry you have active at the time. For the functionality to be configured and stored in the scene file such that end users get the same experience, there would need to be an explicit link there.

PR #496, opened 2026-09-10
balance every loadingManager itemStart with an itemEnd

sdumetz:
`LoadingManager.itemEnd()` should be called after `loadingManager.itemError` if the error is fatal (ie: we have no intention of retrying, fetch a fallback, etc...).

To reproduce the issue, create an audio element pointing at a file that does not exist. Save the scene and reload: The `<sv-spinner>` spins forever. 

The missing audio file isn't visually reported either, but this is another issue.

It was already correct in `ModelReader`, which represent most of the expected failures so we didn't see the unbalance too often (`ModelReader.ts:L152-155`):

```typescript
        .catch((e)=> {
            this.loadingManager.itemError(url);
            throw e;
        }).finally(()=> this.loadingManager.itemEnd( url ));
```

I also fixed the FontReader even if it isn't used anywhere at the moment.

### DIFF
--- a/source/client/io/AudioReader.ts
+++ b/source/client/io/AudioReader.ts
@@ -36,15 +36,17 @@ export default class AudioReader
             request.responseType = 'arraybuffer';
             request.onload = () => {
                 if (request.status >= 200 && request.status < 300) {
-                    this.loadingManager.itemEnd(url)
+                    this.loadingManager.itemEnd(url);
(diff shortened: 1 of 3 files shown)


gjcope:
@sdumetz Thanks for the PR! Merged with [rc-66](https://github.com/Smithsonian/dpo-voyager/tree/rc-66)

